PR #4, opened 2019-05-16
Add template and initialization scripts for kubernetes

brianv0:
This adds a simple initialization script that will resolve a templated version of the authorization config.

The output config files may still need to be modified by the user as appropriate, for example, mapping groups to capabilities.


brianv0:
@athornton would you mind checking this out a bit?

It's pretty close to what you're doing with in nublado, but I pulled in a templating CLI (mustache, via docker) to help out because I was having some weirdness using the simple sed method thanks to base64 encoded values.

We can put this somewhere else if appropriate, hopefully it's structured in a way that's amenable to terraform/kustomize for the future.

athornton:
It looks OK to me.  I'm a bit hesitant on mustache just because it feels like pulling Ruby in if that's the only thing we need it for is a bad idea.  How hard would it be (if we need a real templating framework rather than just sed) would it be to use Jinja2 in its place?

brianv0:
I had just done it that way because there was a docker image with mustache. I see there's a `j2cli` command and a docker image for that, or I can just write a small jinja cli script myself.

athornton:
I would prefer not pulling in Ruby if this is the only reason for it.  Disclaimer: Ruby managed to really get up my nose when I was working at a shop with very conservative change policies and Ruby was all like "you're running 1.8.3.5???  PFFFFT INSTALL 1.8.4.2 YOU MORON...oh, hang on 1.8.4.3 YOU MORON!"  It's probably less obnoxious these days, and "developer-friendly" probably doesn't just mean "operations-hostile", and I'm no longer really in an operations support role, but....

brianv0:
Switched to j2cli (installable with `pip install j2cli[yaml]`) with docker fallback if it's not installed.

PR #8, opened 2019-05-31
Productionize Authorizer Code

brianv0:
This PR brings in code that's running in -int as well as additional project organizational changes to better facilitate unit testing, docker and kubernetes deployment, and updates for moving the repo into the lsst org.

PR #9, opened 2019-05-31
Python-only changes from DM-18834

brianv0:
This PR is to merge only the python changes from DM-18834 into master.

These changes implemented are as follows:
* Move code `authorizer` directory from rootfs/app to root of repo
* Cleanups for mypy, flake8, black
* Implementation of token issuer and token issuer interface (with templates)
* Add Dockerfile multi-stage build
* Reorganize code from __init__.py to app.py and wsgi,py for better uwsgi support and to enable unit testing
* Doc fixes in preparation of sphinx autohttp doc generation
* Different ticket issuance and encryption strategy based on changes to oauth2_proxy


brianv0:
For this review, I think I'd like to feed back any changes back into tickets/DM-18834 via cherry picks, but I'm keeping it open as PR against master in any case.

authorizer/authnz.py
```diff
@@ -0,0 +1,218 @@
...
+    successes = []
+    message = ""
+    for check_access in check_access_callables:
+        logger.debug(f"Checking access using {check_access.__name__}")
+        (successful, message) = check_access(capability, verified_token)
+        if not successful:
+            break
+        successes.append(successful)
+
+    success = sum(successes) == len(check_access_callables)
```
athornton: /blink /blink oh, I get it.  Cool.

authorizer/tokens.py
```diff
@@ -0,0 +1,465 @@
...
+    if "." not in trimmed_ticket:
+        print(2)
+        return None
+    ticket_id, secret_b64 = trimmed_ticket.split(".")
+    try:
+        int(ticket_id, 16)  # Check hex
+        secret = base64.urlsafe_b64decode(add_padding(secret_b64))
+        return Ticket(ticket_id=ticket_id, secret=secret)
+    except (ValueError, Error) as e:
+        print(e)
```
athornton: Should this be a print, or some kind of logging statement?

requirements.txt
```diff
@@ -9,3 +9,5 @@ dynaconf
 redis
 flask_wtf
```
athornton: These are some great module names here.

authorizer/app.py
```diff
@@ -0,0 +1,364 @@
...
+    response.headers["X-Auth-Request-Capabilities-Satisfy"] = satisfy
+
+
+def _make_success_headers(response: Response, encoded_token: str) -> None:
+    """Set Headers that will be returned in a successful response.
+    :return: The mutated response object.
+    """
+    _make_capability_headers(response, encoded_token)
+
+    decoded_token = jwt.decode(encoded_token, verify=False)
```
athornton: Why are we not verifying here?  Is it that we assume it's already been done?

brianv0: yeah, `_make_success_headers` should only be called if we are successful with authentication. 

I can add an extra arg to the function which accepts the verified token _also_ instead of decoding it again (same with `_make_capability_headers`). I think the lack of that argument is mostly a relic from before we were reissuing (and just passing the encoded token along)

authorizer/authnz.py
```diff
@@ -0,0 +1,218 @@
...
+        success = sum(successes) == len(capabilities)
+    message = ", ".join(messages)
+    return success, message
+
+
+def check_authorization(capability: str, verified_token: Mapping[str, Any]) -> Tuple[bool, str]:
+    """
+    Check the authorization for a given capability.
+    A given capability may be authorized by zero, one, or more criteria,
+    modeled as a callables. All callables MUST pass, returning True,
```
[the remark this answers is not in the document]
brianv0:
```suggestion
    modeled as callables. All callables MUST pass, returning True,
```

authorizer/app.py
```diff
@@ -0,0 +1,364 @@
...
+    original_auth = _find_token(ORIGINAL_TOKEN_HEADER) or ""
+    oauth2_proxy_ticket = original_auth if original_auth.startswith(f"{ticket_prefix}:") else ""
+    reissue_requested = request.args.get("reissue_token", "").lower() == "true"
+    if reissue_requested:
+        encoded_token, oauth2_proxy_ticket = _check_reissue_token(encoded_token, decoded_token)
+    response.headers["X-Auth-Request-Token"] = encoded_token
+    response.headers["X-Auth-Request-Token-Ticket"] = oauth2_proxy_ticket
+
+
+def _check_reissue_token(encoded_token: str, decoded_token: Mapping[str, Any]) -> Tuple[str, str]:
```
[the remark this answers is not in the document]
brianv0: How about something like this?

```
def _check_reissue_token(encoded_token: str, decoded_token: Mapping[str, Any]) -> Tuple[str, str]:
    ...
    new_audience = None

    if not from_this_issuer:
        new_audience = current_app.config.get("OAUTH2_JWT.AUD.DEFAULT", "")
        ticket = parse_ticket(cookie_name, oauth2_proxy_ticket_str)
        # If we didn't issue it, it came from a provider, and it is
        # inherently a new session, and this happens only once, after
        # initial login. If there's no cookie, or we failed to
        # parse it, there's something funny going on.
        assert ticket, "ERROR: OAuth2 Proxy cookie must be present"
    elif from_this_issuer and from_default_audience and to_internal_audience:
        # In this case, we only reissue tokens from a default audience
        new_audience = current_app.config.get("OAUTH2_JWT.AUD.INTERNAL", "")
        ticket = Ticket()

    if new_audience:
        encoded_token = issue_token(decoded_token, new_audience, True, ticket)
        oauth2_proxy_ticket_str = ticket.encode(cookie_name)
    return encoded_token, oauth2_proxy_ticket_str
```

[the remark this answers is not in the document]
brianv0: I went ahead and changed it according to that, I was able to drop about 40 LoC in the process.

brianv0:
These changes have been cherry-picked back to #8.

PR #10, opened 2019-06-18
Fixups for new oauth2_proxy cookie/session handling

brianv0:


brianv0:
This includes several fixes:
- new `oauth2_proxy` changes changed legacy session serialization in a breaking manner. We switch to the "new" serialization which is JSON based. There were other minor fixes around encryption/decryption handling related to this.
- `oauth2_proxy` switched to a base64-encoded, signed cookie, so tickets need to be base64 decoded from that cookie.
- Kubernetes config fixes
- Documentation fixes

PR #15, opened 2020-03-12
Refactor build system and add /analyze endpoint

rra:
- Modernize the build system to match the roundtable_aiohttp_bot template
- Fix the existing test
- Add tests for the issue_token path and some related functions
- Add a new /analyze route to show information about a ticket or token


src/jwt_authorizer/analyze.py
```diff
@@ -0,0 +1,80 @@
...
+    output: Dict[str, Any] = {
+        "ticket": {
+            "ticket_id": ticket.ticket_id,
+            "secret": base64.urlsafe_b64encode(ticket.secret).decode(),
+        }
+    }
+
+    session = token_store.get_session(ticket)
+    if not session:
+        output["errors"] = ["No session found for {ticket.as_handle()}"]
```
ktlim: Does this need an `f` prefix or is it interpreted somewhere else?

rra: Good catch, thanks!

rra:
BTW, this commit introduces a fair amount of duplicate code because I pulled out some things into objects that can use dependency injection to make tests easier. Future refactorings will reconsolidate.

src/jwt_authorizer/tokens.py
```diff
@@ -91,15 +97,23 @@ def parse_ticket(prefix: str, ticket: str) -> Optional[Ticket]:
     ticket_id, secret_b64 = trimmed_ticket.split(".")
     try:
         int(ticket_id, 16)  # Check hex
-        secret = base64.urlsafe_b64decode(add_padding(secret_b64))
+        secret = base64.b64decode(
+            add_padding(secret_b64), altchars=b"-_", validate=True
+        )
+        if secret == b"":
+            raise ValueError("ticket secret is empty")
```
cbanek: This looks slightly different but I'm not sure how it's really different?  Seems legit either way?

src/jwt_authorizer/tokens.py
```diff
@@ -91,15 +97,23 @@ def parse_ticket(prefix: str, ticket: str) -> Optional[Ticket]:
...
         return Ticket(ticket_id=ticket_id, secret=secret)
     except (ValueError, Error) as e:
-        logger.info("Error decoding ticket:", e)
+        logger.info("Error decoding ticket: %s", str(e))
```
cbanek: Does this happen a lot?  While it's good to not spend a lot of cycles with bad auth, I wonder if we might want to bump this up to error level?

cbanek: Or I guess this is one of those functions that is called a lot but failure is normal?

.gitignore
```diff
@@ -1,11 +1,134 @@
...
+# mypy
+.mypy_cache/
+.dmypy.json
+dmypy.json
+
+# Pyre type checker
+.pyre/
 
+# pytype static type analyzer
+.pytype/
```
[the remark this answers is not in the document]
rra: Jonathan has been using a stock one for all his packages, so I just copied it. :)

Dockerfile
```diff
@@ -1,36 +1,69 @@
...
+#   stages.
+# install-image
+#   Installs the app into the virtual environment.
+# runtime-image
+#   - Copies the virtual environment into place.
+#   - Runs a non-root user.
+#   - Sets up the entrypoint and port.
 
-RUN pip install mypy
+FROM python:3.7-slim-buster as base-image
```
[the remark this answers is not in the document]
rra: Yeah, this too is from Jonathan's work based on a bunch of blog posts for speeding things up. It's pretty awesome.

Dockerfile
```diff
@@ -1,36 +1,69 @@
...
+# Add configuration
 ADD rootfs/ /
-ADD . /app
 
-EXPOSE 8080
-ARG COMMIT=""
-ARG COMMIT_DESCRIBE=""
-ARG BRANCH=""
+# Make sure we use the virtualenv
+ENV PATH="/opt/venv/bin:$PATH"
```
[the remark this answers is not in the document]
rra: I believe so, although I'm just copying from the roundtable_aiohttp_bot template.

kube/template/deployment/oauth2-proxy.yml
```diff
@@ -19,28 +19,28 @@ spec:
         imagePullPolicy: Always
         name: oauth2-proxy
         args:
```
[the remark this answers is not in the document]
rra: This is deprecated in favor of the Helm chart and will hopefully be deleted entirely. The changes are just to make the YAML linter not blow up on everything until I can delete them entirely.

src/jwt_authorizer/tokens.py
rra: The `validate=True` is the difference. Without that, Python silently discards invalid base64 characters, which I don't think we want to do here. Unfortunately, there's no `validate` flag to `urlsafe_b64decode` so I have to use the generic function and set `altchars`.

cbanek: Sounds good!

src/jwt_authorizer/tokens.py
rra: Thanks, good catch. Bumping the logging level.

rra:
Yeah, there's a ton of duplicate code in the tests right now. I'm definitely going to fix that -- just thinking about how. I'm trying to avoid a pattern where the test doesn't make sense by itself because you need to know a bunch of settings that are set off somewhere else, but there's definitely too much boilerplate right now.

PR #24, opened 2020-04-25
[DM-23943] Use pydantic for configuration files

rra:
- Replace Dyanconf with pydantic
- Add validation of URLs
- Add syntax checking of the example configuration files and split GitHub and OIDC examples


src/gafaelfawr/config.py
```diff
@@ -168,41 +215,49 @@ class Config:
...
+    known_scopes: Mapping[str, str]
     """Known scopes (the keys) and their descriptions (the values)."""
 
-    username_claim: str
-    """Token claim from which to take the username."""
-
-    uid_claim: str
-    """Token claim from which to take the UID."""
-
     safir_config: SafirConfig
```
jonathansick, line 230: Rename as `safir` so config is not as redundant when accessing the attribute?

rra:
Yeah, I was not entirely happy with the duplication into a pydantic model either. There were three things that pushed me to do it that way:

- I wanted to have self-contained configuration for each component, which required duplicating some settings, and that seemed tricky to do inside pydantic's hooks (although doable).
- In a couple of places the structure is different inside the program than outside (`group_mapping` is at the top level in the configuration but is only part of the issuer config, `group_mapping` is laid out differently than it is in the configuration file).
- There are multiple configuration settings that are post-processed to create new objects internally: all the `_file` secrets are read from disk during configuration, and the RSA private key is converted to an `RSAKeyPair` object.

I think it would be possible to do all of this with pydantic hooks, but it might be messier. I don't think this code is the final state or last word and we should reconsider it and see if we can come up with a better way of deduplicating that code.

src/gafaelfawr/config.py
rra: Good idea. Will do.

PR #25, opened 2020-04-27
[DM-24302] Avoid authentication redirects for AJAX requests

rra:
- Don't redirect AJAX requests to external authentication providers
- Clean up and refactor some of the login code to make it easier to follow
- Hide the authentication provider implementation from the login route entirely


rra:
This one is probably easier to review by looking at each commit in isolation, since a couple of the commits are just moving code around.

PR #26, opened 2020-04-27
Add badges for GitHub build and Docker image

rra:


rra:
Good call. I've updated the alt text.

PR #32, opened 2020-05-13
[DM-24810] Add support for logging the real client IP

rra:
Using aiohttp_remotes to trust X-Forwarded-For is sufficient to log
the true client IP in a normal NGINX ingress deployment for all the
routes other than /auth, but because /auth is an auth_request
subrequest handler, it still logged the IP address of the NGINX
ingress.

To avoid this, document how to configure the NGINX ingress to
construct a full chained X-Forwarded-For header instead of replacing
it at each hop, and add a new configuration directive, proxies, that
tells Gafaelfawr what IP addresses are internal to the Kubernetes
cluster so that it can walk backwards and find the true client IP.

Also document the need to disable source NAT (particularly in GKE)
so that the NGINX ingress will see the true client IP.


docs/install.rst
```diff
@@ -171,7 +176,7 @@ The typical annotations for a web application used via a web browser are:
 
    annotations:
     kubernetes.io/ingress.class: nginx
-    nginx.ingress.kubernetes.io/auth-request-redirect: $request_uri
+    nginx.ingress.kubernetes.io/auth-method: GET
```
cbanek, line 179: why is this one changing?

docs/logging.rst
```diff
@@ -65,3 +66,55 @@ The ``/login`` route adds the following attributes:
...
+For the curious, here are the details of why these changes are required.
+
+Determining the client IP from ``X-Forwarded-For`` is complicated because Gafaelfawr's ``/auth`` route is called via an NGINX ``auth_request`` directive.
+In the Kubernetes NGINX ingress, this involves three layers of configuration.
+The protected service will have an ``auth_request`` directive that points to a generated internal location.
+That internal location will set ``X-Forwarded-For`` and then proxy to the ``/auth`` route.
+The ``/auth`` route configuration is itself a proxy that also sets ``X-Forwarded-For`` and then proxies the request to Gafaelfawr.
+Because of this three-layer configuration, if NGINX is configured to always replace the ``X-Forwarded-For`` header, Gafaelfawr will receive a header containing only the IP address of the NGINX ingress.
+
+The above configuration tells the NGINX ingress to instead retain the original ``X-Forwarded-For`` and append each subsequent client IP.
```
cbanek, line 115: I just noticed that I'm not sure this says what header the original host is emitted to?  Or are we not going to pass that back to people as well?

docs/install.rst
rra: Yeah, sorry, this is somewhat unrelated and I didn't update the PR overview. It turns out that the `auth-request-redirect` that we were setting is just the default and thus doesn't do anything, so removing it makes everything shorter. And the `auth-method` setting does two things: first, if the attempt is a `POST`, apparently it would have sent a `POST` auth request, which would break everything. This overrides that. And second it will force NGINX to set `X-Original-URI`, which we wanted for logging.

docs/logging.rst
rra: Oh, that is added as `X-Original-Host`. I'll note that explicitly.

rra: `X-Forwarded-Host` as it turns out. But I think I misunderstood your note -- the hostname of the client isn't recorded or logged anywhere. `X-Forwarded-Host` is the `Host` header sent by the client and names the host of the server. That isn't logged by Gafaelfawr, although it does use it for the `/login` and `/logout` routes. I'll make a note of that.

docs/install.rst
cbanek: is the change to auth-method something we need to apply to all the ingress rules?

docs/logging.rst
cbanek: Sorry yeah this is the confusing part about which host - sorry I meant the client IP of the original machine making the request - which header is that emitted to?  Is that something a backend service will have access to to log itself as well? Or is it all just logged in gafaelfawr  and not passed on?

Not so worried about the original Virtual host field thing, which is the other Host field that I always get confused with :)

docs/install.rst
rra: Yup, it is. I'll make another PR for lsp-deploy to set the `auth-method` annotation.

docs/logging.rst
rra: Oh, the client IP of the original machine ends up in `X-Forwarded-For`. Gafaelfawr only parses it, and doesn't set the results of that parse into any supplemental header, but the protected service would receive the same `X-Forwarded-For` header and could in theory do the same thing.

Do you think we should set the results of Gafaelfawr's parsing in some other header that the backend could use more simply?

cbanek: Ah okay.  I think it's fine for now, I forgot that it would probably be there on the other side as well.  But sometimes frameworks also tend to mess this up and not deal with picking the right one, as you well know having fixed one.  It's probably fine for now, although I could maybe see just taking what we know to be the real IP and plopping that somewhere, but that's probably a future thing.

PR #34, opened 2020-05-29
[DM-24930] Gafaelfawr cleanups

rra:
- Add a request context so that the auth method can be added to the logs
- Remove unnecessary debug logging by making info-level logging more useful
- Improve test coverage for the session store and token verifier
- Format `/auth/analyze` output
- Add `pytest-sugar`
- Support configuration overrides via `create_app`
- Move `XForwardedFiltered` to its own file
- Separate logging tests
- Expose IP address of client in auth headers
- Refresh dependencies
- Better error message when unable to reach the auth provider
- Drop support for Python 3.7
- Refactor `ClientSession` mocks


rra:
This is a bit of a grab bag, sorry. Probably best reviewed commit by commit instead of all at once.

src/gafaelfawr/verify.py
```diff
@@ -337,12 +342,12 @@ async def _get_keys(self, issuer_url: str) -> List[Dict[str, str]]:
             raise FetchKeysException(msg)
 
         body = await r.json()
-        if "keys" not in body:
+        try:
+            return body["keys"]
+        except Exception:
```
cbanek, line 347: I think I liked the pattern before with looking at the key rather than catching exception, but it's fine either way (not sure if I'm missing some subtle detail)

rra: This is super, super-fiddly and probably won't come up in practice, but I made this change because I realized that if upstream returned a JSON list with `keys` as one of the members of the list, it would pass the if statement and then produce an uncaught exception. This way of writing it ensures we catch the exception so that we can report an error message to the user.

cbanek: Sounds good, that is hilariously subtle.

PR #93, opened 2020-07-30
[DM-26072] Refactor in preparation for adding an OpenID Connect server

rra:
- Add an abstraction layer for the encrypted Redis store so that it can be reused for OpenID Connect
- Rename the `TokenStore` class and related things to be a bit clearer
- Move all exceptions into a single module rather than scattering them through the code base
- Add a helper function for tests that need session authentication
- Avoid empty `isMemberOf` in generated GitHub tokens
- Switch the test suite to aioresponses


rra:
This is kind of a tedious review since it's all refactoring. Each commit should be reasonably self-contained. The need for some of this in order to share code (such as the Redis store refactoring) will become obvious in the OpenID Connect server PR.

src/gafaelfawr/exceptions.py
```diff
@@ -0,0 +1,78 @@
...
+class MissingClaimsException(VerifyTokenException):
+    """The token is missing required claims."""
+
+
+class UnknownAlgorithmException(VerifyTokenException):
+    """The issuer key was for an unsupported algorithm."""
+
+
+class UnknownKeyIdException(VerifyTokenException):
+    """The reqeusted key ID was not found for an issuer."""
```
cbanek, line 88: I honestly love this pattern, having all the errors in the same place really gives good context and useful info!

cbanek:
Sorry this took so long - this all looks great!

rra:
Thank you for the review!

PR #95, opened 2020-08-07
[DM-26072] Add an OpenID Connect server

rra:
This adds a simple OpenID Connect server that should be sufficient for Chronograf.
It also includes some supporting refactoring changes:

- Helper function for generating `WWW-Authenticate` challenges
- Merge `Authorization` header parsing
- Centralize the route decorators in one module
- Add a helper function to rebind the request logger
- Refactor verification of a return URL

### DIFF
--- a/docs/_static/flow-oidc.diag
+++ b/docs/_static/flow-oidc.diag
@@ -1,18 +1,18 @@
 seqdiag {
     browser -> app;
-    browser <- app [label = "redirect to /auth/openid/login"];
+    browser <- app [label = "redirect"];
     browser -> gafaelfawr [label = "/auth/openid/login"];
-    browser <- gafaelfawr [label = "redirect to /login"]
+    browser <- gafaelfawr [label = "redirect"]
     browser -> gafaelfawr [label = "/login"];
-    browser <- gafaelfawr [label = "redirect to IdP"];
+    browser <- gafaelfawr [label = "redirect"];
     browser -> IdP [label = "authenticate"];
     browser <- IdP [label = "redirect to /login"];
     browser -> gafaelfawr [label = "/login"];
                gafaelfawr -> IdP [label = "get data"];
                gafaelfawr <- IdP [label = "user data"];
-    browser <- gafaelfawr [label = "redirect to /auth/openid/login"];
+    browser <- gafaelfawr [label = "redirect with session cookie"];
     browser -> gafaelfawr [label = "/auth/openid/login"];
-    browser <- gafaelfawr [label = "redirect to app"];
+    browser <- gafaelfawr [label = "redirect"];
     browser -> app;
                app -> gafaelfawr [label = "/auth/openid/token"];
                app <- gafaelfawr;
--- a/docs/_static/flow-oidc.diag
+++ b/docs/_static/flow-oidc.diag
@@ -1,0 +1,22 @@
+seqdiag {
+    browser -> app;
+    browser <- app [label = "redirect to /auth/openid/login"];
+    browser -> gafaelfawr [label = "/auth/openid/login"];
+    browser <- gafaelfawr [label = "redirect to /login"]
+    browser -> gafaelfawr [label = "/login"];
+    browser <- gafaelfawr [label = "redirect to IdP"];
+    browser -> IdP [label = "authenticate"];
+    browser <- IdP [label = "redirect to /login"];
+    browser -> gafaelfawr [label = "/login"];
+               gafaelfawr -> IdP [label = "get data"];
+               gafaelfawr <- IdP [label = "user data"];
+    browser <- gafaelfawr [label = "redirect to /auth/openid/login"];
+    browser -> gafaelfawr [label = "/auth/openid/login"];
+    browser <- gafaelfawr [label = "redirect to app"];
+    browser -> app;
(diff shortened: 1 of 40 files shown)


rra:
This one is probably best reviewed in one lump, since the patch sequence has some false starts on how best to express some ideas that I backed out of later.

docs/_static/flow-oidc.diag
```diff
@@ -0,0 +1,22 @@
...
+    browser -> gafaelfawr [label = "/login"];
+    browser <- gafaelfawr [label = "redirect to IdP"];
+    browser -> IdP [label = "authenticate"];
+    browser <- IdP [label = "redirect to /login"];
+    browser -> gafaelfawr [label = "/login"];
+               gafaelfawr -> IdP [label = "get data"];
+               gafaelfawr <- IdP [label = "user data"];
+    browser <- gafaelfawr [label = "redirect to /auth/openid/login"];
+    browser -> gafaelfawr [label = "/auth/openid/login"];
+    browser <- gafaelfawr [label = "redirect to app"];
```
cbanek, line 15: is there a cookie in there or something?  This is super helpful though

src/gafaelfawr/handlers/oidc.py
```diff
@@ -0,0 +1,229 @@
...
+    context = RequestContext.from_request(request)
+    data = await request.post()
+    grant_type = get_form_value(data, "grant_type")
+    client_id = get_form_value(data, "client_id")
+    client_secret = get_form_value(data, "client_secret")
+    code = get_form_value(data, "code")
+    redirect_uri = get_form_value(data, "redirect_uri")
+
+    # Check the request parameters.
+    if not grant_type or not client_id or not code or not redirect_uri:
```
cbanek, line 186: if not all([grant_type, client_id, code, redirect_uri])?

tests/handlers/oidc_test.py
```diff
@@ -0,0 +1,506 @@
...
+        "remote": "127.0.0.1",
+        "request_id": ANY,
+        "user_agent": ANY,
+    }
+
+
+async def test_login_errors(
+    create_test_setup: SetupTestCallable, caplog: LogCaptureFixture
+) -> None:
+    clients = [OIDCClient(client_id="some-id", client_secret="some-secret")]
```
cbanek, line 192: since you use some-id and some-secret a lot, maybe just make it a constant?  that way it's a bit easier to see where you use it in the code, since strings don't stick out as much

cbanek: Or maybe just pull them out?  I'm not sure where you override it

tests/support/headers.py
```diff
@@ -0,0 +1,78 @@
...
+from gafaelfawr.handlers.util import AuthChallenge, AuthError, AuthType
+
+if TYPE_CHECKING:
+    from typing import Dict, List
+
+
+def parse_www_authenticate(header: str) -> AuthChallenge:
+    """Parse a ``WWW-Authenticate`` header into this representation.
+
+    A ``WWW-Authenticate`` header consists of one or mor challenges, each of
```
cbanek, line 18: more

tests/support/headers.py
```diff
@@ -0,0 +1,78 @@
...
+    # be match group 2.
+    #
+    # Each attribute has to either start at the beginning of the portion of
+    # the header after the auth type (\A) or follow a previous attribute with
+    # a comma and whitespace (,\s*), ensuring there isn't any extraneous junk
+    # in the header.
+    error = None
+    error_description = None
+    scope = None
+    for attribute in re.finditer(r'(?:\A|,\s*)([^ "=]+)="([^"]+)"', info):
```
cbanek, line 41: yum

docs/_static/flow-oidc.diag
rra: Yeah, `/login` sets a cookie. I have now made that more explicit. The redirect back from the app includes the code in the URL as a GET parameter.

src/gafaelfawr/handlers/oidc.py
rra: Oh, good idea. I hadn't thought of that trick.

tests/handlers/oidc_test.py
rra: Hm, they're just arbitrary strings that we can check match in other places later on. There are no default values, so it's not overridden; if the test doesn't set them, there would be no valid clients. They feel the same as strings like `"githubuser"` or `"random-state"`; they're data that we're testing is persisted or checked properly through the code.

But I'm not sure if that's what you're asking, or if I'm correctly envisioning the alternative you're thinking of? (I agree that this test is long and tedious to read, and it's not totally clear what's being checked, but I wasn't sure how to make it better. It's a fairly mechanical exploration of the error paths.)

cbanek: Ah I gotcha that makes sense, yeah it is a bit mechanical, but I guess that's fine

src/gafaelfawr/handlers/oidc.py
rra: Sadly, mypy hadn't thought of that trick either, and it causes type checking to not work because it doesn't realize that these variables can no longer be `None`.

PR #107, opened 2020-09-11
[DM-26682] Fix OpenID Connect bugs

rra:
- Fix type of `expires_in` data element returned by `/auth/openid/token`
- Fix encoding of JWKS


afausti (approved):
LGTM!

afausti (approved):
Looks great! I've just tested the latest fix and I was able to login to Chronograf.

PR #109, opened 2020-09-16
[DM-26682] Fix various issues with InfluxDB tokens

rra:
- Put the username in `username`, not `sub`, for InfluxDB tokens
- Add a new configuration parameter to force the username to a single value


afausti (approved):
Thanks, I've tested these changes and they work great.

PR #119, opened 2020-10-26
[neophile] Update dependencies

sqrbot:
- Update frozen Python dependencies
- Update pre-commit/pre-commit-hooks pre-commit hook from v3.2.0 to v3.3.0


rra:
Fixed as part of #120

PR #120, opened 2020-10-27
[DM-27009] Convert Gafaelfawr to FastAPI

rra:
Convert from aiohttp to FastAPI and httpx.


rra:
This is very long and mostly shuffling equivalent code around and changing implementation details without changing behavior, so I recommend skimming rather than trying to read line by line. Let me know if there are any details I can fill in.

jonathansick:
Great, I'll take a look.

tests/conftest.py
```diff
@@ -41,106 +57,52 @@ def driver() -> Iterator[webdriver.Chrome]:
...
-
-    Parameters
-    ----------
-    tmp_path : `pathlib.Path`
-        Path to a per-test temporary directory, injected as a fixture.
-    responses : `aioresponses.aioresponses`
-        Mock object for `aiohttp.ClientSession` call handling.
-    aiohttp_client : `typing.Callable`
-        Function creates an aiohttp test client.
+async def setup(
```
jonathansick, line 89: Something that's common in other applications I've worked with is that the test client fixture is responsible for setting up the application instance. That way you don't need to have two fixtures for every test.

On the other hand, maybe this approach is more transparent and flexible? I know that for some apps I'd need to create multiple client fixtures to have different application set ups (like a client that has different roles assigned to it).

Since the `client` fixture already depends on the `setup` fixture, I wonder if a test _could_ actually just use the `client` fixture in cases where it doesn't use the `setup` directly (i.e., is the docstring below still correct?).

jonathansick (approved):
I like this. The app is well structured and its easy to find where things are. I think the dependency architecture works out pretty well seeing it implemented.

- How would you host the Gafaelfawr at a sub-path? Or is gafaelfawr always hosted at the root?

- It's neat that you're able to separate the settings file schema validation from the internal configuration dataclasses. I wonder if it might be possible to package them up so they're a little more related? Maybe a `config` subpackage, with a `settingsschema.py` module in it? Having a `config.py` and a `settings.py` both at the root of the app was confusing at first glance.

tests/support/setup.py
```diff
@@ -0,0 +1,377 @@
...
+    from pytest_httpx import HTTPXMock
+    from pytest_httpx._httpx_internals import Response
+
+    from gafaelfawr.config import Config, OIDCClient
+    from gafaelfawr.keypair import RSAKeyPair
+    from gafaelfawr.providers.github import GitHubUserInfo
+    from gafaelfawr.tokens import Token, VerifiedToken
+
+
+class SetupTest:
```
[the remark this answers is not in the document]
rra: Unfortunately, the problem is that pytest will then think it's a test case and try to execute it. I've added a documentation note to that effect.

tests/conftest.py
rra: What the docstring is getting at is that `SetupTest` is responsible for creating the configuration, and until the configuration is created, attempting to create a client wrapping the ASGI application will immediately fail because it will try to run the lifecycle manager on the ASGI app, which will trigger the startup event, which will try to read the configuration file, and that will fail because `/etc/gafaelfawr/gafaelfawr.yaml` doesn't exist.

I'm going to try to turn `SetupTest` into an async context manager so that it can do both its own setup and that of the client and eliminate one of the fixtures. Stay tuned....

tests/support/setup.py
```diff
@@ -0,0 +1,377 @@
...
+        self,
+        *,
+        tmp_path: Path,
+        config: Config,
+        redis: Redis,
+        httpx_mock: HTTPXMock,
+    ) -> None:
+        self.tmp_path = tmp_path
+        self.config = config
+        self.http_client = AsyncClient()
```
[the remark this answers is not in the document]
rra: Nope, this is a good catch. I have now done this as part of the refactoring to make the setup fixture responsible for configuring everything.

tests/conftest.py
rra: I think that worked! Good idea.

rra:
Merging `config.py` and `settings.py` is a good idea. Now done. Gafaelfawr currently doesn't support subpaths (a possibly poor decision on my part), so I haven't bothered to try to figure out how to make that work the way that it did with aiohttp. It doesn't look like FastAPI has any equivalent functionality to ask what the path to a route is, so I'd need to write something, and I was going to defer that until we had a need.

PR #125, opened 2020-11-12
[DM-27342] Remove internal JWTs, add first pass of token API

rra:
- Add the user-facing portion of the token API, without the history support
- A stub of admin support is there because it was added as a test, but it isn't usable yet
- Remove internal JWTs and the old session mechanism in favor of the new opaque tokens
- Add a database with the full schema from SQR-049
- Add support for internal and notebook tokens and the new SQR-049 method of requesting delegated tokens

Known bugs to be addressed in later PRs:

- The mechanism for injecting the database URL will need to be more complex
- Switch to the FastAPI Docker container base and update the test version of the application
- Test with PostgreSQL as well as SQLite
- Handle child tokens of tokens that have changed expiration
- Flesh out and sanity-check the generated OpenAPI documentation


rra:
Just rebased against current master branch.

src/gafaelfawr/services/token.py
```diff
@@ -0,0 +1,473 @@
...
+            Whether the token was found and deleted.
+        """
+        info = self.get_info(key, username)
+        if not info:
+            return False
+        if info.username != auth_data.username:
+            msg = f"Token owned by {info.username}, not {auth_data.username}"
+            self._logger.warning("Permission denied", error=msg)
+            raise PermissionDeniedError(msg)
+        await self._token_redis_store.delete(key)
```
ktlim, line 189: Are there any potential race conditions by having this removed from Redis before Postgres?  (Or vice versa for creation, where it appears in Redis before Postgres.)

src/gafaelfawr/services/token.py
```diff
@@ -0,0 +1,473 @@
...
+            raise PermissionDeniedError("Token does not have required scopes")
+
+        # See if there's already a matching internal token.
+        key = self._token_db_store.get_internal_token_key(
+            token_data, service, scopes
+        )
+        if key:
+            data = await self._token_redis_store.get_data_by_key(key)
+            if data:
+                return data.token
```
ktlim, line 267: I assume the reissue-if-half-expired check is yet to be implemented.

src/gafaelfawr/storage/token.py
```diff
@@ -0,0 +1,383 @@
...
+        if token_name:
+            name_conflict = (
+                self._session.query(SQLToken.token)
+                .filter_by(username=data.username, token_name=token_name)
+                .scalar()
+            )
+            if name_conflict:
+                msg = f"Token name {token_name} already used"
+                raise DuplicateTokenNameError(msg)
+        new = SQLToken(
```
ktlim, line 78: If there's a race, you'll miss out on raising `DuplicateTokenNameError` for the unique key violation, right?

ktlim (approved):

docs/arch/providers.rst
```diff
@@ -26,22 +24,13 @@ Refresh tokens are not used.
...
-``sub``
-    The ``login`` attribute returned by the ``/user`` API route, forced to lowercase.
-``uid``
-    The ``login`` attribute returned by the ``/user`` API route, forced to lowercase.
-``uidNumber``
-    The ``id`` attribute returned by the ``/user`` API route, converted to a string.
-    The hope is that this is suitable for a unique UID.
+GitHub does not issue JWTs, so the token created after GitHub authentication is based on information retrieved from the GitHub API.
+The username will be taken from the ``login`` value returned by the ``/user`` API route, forced to lowercase.
+The UID will be taken from the ``id`` value returned by the ``/user`` API route.
```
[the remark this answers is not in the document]
rra: Correct. I added a note about that.

docs/arch/routes.rst
```diff
@@ -38,20 +42,10 @@ Gafaelfawr supports the following routes:
     Retrieves a JWT given an OpenID Connect authorization code obtained via an authentication request.
     The parameters to this route are those for the Token Request in the Authorization Code Flow in `OpenID Connect`_.
 
-``/auth/tokens``
-    Displays all user-issued tokens for the authenticated user.
-
 ``/auth/tokens/influxdb/new``
```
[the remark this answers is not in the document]
rra: The code for that has been temporarily deleted because it was incompatible with the new token scheme, but will be replaced with the new UI that I'm currently working on before the new Gafaelfawr is deployed anywhere.

src/gafaelfawr/handlers/api.py
```diff
@@ -0,0 +1,241 @@
...
+    response_model=TokenInfo,
+    response_model_exclude_none=True,
+)
+async def get_token(
+    username: str,
+    key: str,
+    auth_data: TokenData = Depends(authenticate_session),
+    context: RequestContext = Depends(context_dependency),
+) -> TokenInfo:
+    if username != auth_data.username:
```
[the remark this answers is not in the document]
rra: Yeah, agreed, it's a bit of a layering violation because I was trying too hard for code reuse. I'll add a new service layer method that does the authorization check instead.

Correct, admin impersonation isn't implemented yet. Sorry, I should have noted that explicitly.

docs/arch/routes.rst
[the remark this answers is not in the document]
ktlim: Sorry, maybe it wasn't obvious that the reference was to the InfluxDB tokens.

src/gafaelfawr/schema/subtoken.py
```diff
@@ -0,0 +1,23 @@
...
+
+from gafaelfawr.schema.base import Base
+
+__all__ = ["Subtoken"]
+
+
+class Subtoken(Base):
+    __tablename__ = "subtoken"
+
+    id = Column(Integer, primary_key=True)
```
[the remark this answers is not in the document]
rra: Yup, just now fixed this to make `child` the primary key.

src/gafaelfawr/services/token.py
rra: Redis is canonical and PostgreSQL is secondary, so it's harmless for it to appear in PostgreSQL but not Redis (and will eventually be cleaned up by the as-yet-to-be-written housekeeping job). Good point on creation; I've now reversed the order so that the token is created in PostgreSQL first and then in Redis, on the principle that the PostgreSQL orphans are the harmless ones.

src/gafaelfawr/services/token.py
rra: Correct, along with the internal cache.

src/gafaelfawr/storage/token.py
rra: This happens inside a transaction, so hopefully with most databases that will make the race at least unlikely, but yes, it's possible that some races will result in an uncaught SQL exception. If you know of a good technique to avoid those in general, I'd love to hear more; I've never found a great way of doing that and tend to live with a few narrow race conditions producing 500 errors in database-driven web applications if two servers race the creation of an object in just the right way.

docs/arch/routes.rst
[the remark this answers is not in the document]
rra: Oh! Yes, this will eventually be incorporated into the token API, but a bit farther down the road. Meanwhile, we're using it in a few specific places, so I left the old method in place for compatibility.

rra:
Thank you for the code review!

src/gafaelfawr/storage/token.py
ktlim: In this case, I don't think a race condition would cause database corruption or even improper behavior; it's only that the error returned could be less descriptive/friendly than it would be otherwise.  So no big deal.

PR #129, opened 2020-12-03
[DM-27700] Fix Docker and local dev server for FastAPI

rra:
- Use the recommended FastAPI Docker image as a base
- Fix `tox -e run` to start the server properly with PostgreSQL
- Add an example Docker Compose configuration to run the whole application locally in a Docker container to test the `Dockerfile`


jonathansick:
I only had a brief initial look tonight; I'll read through more tomorrow. I noticed that in the new docker set up the package is no longer being "installed," right, since the source directory is being copied over? So setuptools_scm won't provide version information? Or am I reading the Dockerfile wrong?

Dockerfile
```diff
@@ -14,50 +14,26 @@
 #   - Runs a non-root user.
 #   - Sets up the entrypoint and port.
 
-FROM python:3.8-slim-buster as base-image
+FROM tiangolo/uvicorn-gunicorn-fastapi:python3.8 as base-image
```
jonathansick, line 17: My concern here is that this sort of duplicates how the fastapi version is managed since this image is setting the version of fastapi. Or is the idea that the pip installation of `requirements.txt` will upgrade or downgrade fastapi as necessary? Part of me doesn't like that the docker image is managing a bunch of Python dependencies so we lose the consistency of installing into a clean environment with just the packages that we compiled with pip-tools; I don't know if that's a practical concern?

rra: It's definitely a concern, and you're right that I'm not sure how pip affects that. I hadn't thought that part through and need to go read the upstream Dockerfile. What it gains is that it handles all the complicated setup for uvicorn and gunicorn automatically. Maybe we should talk about the trade-off at co-work?

rra:
Oh, that's right, I forgot about that. Yeah, it insists on having the source in `/app/app`, so I'm not sure how to handle that and make scm work properly.

Dockerfile
jonathansick: Yeah, let's chat about this at co-work. I'm not opposed to moving to a different base image; I just want to be sure we understand what we gain, lose, and how things work in the new system.

rra:
Okay, take a look at this iteration. I think I got it back to substantially the same pattern as we were previously using, but still using the upstream setup machinery.

Dockerfile
```diff
@@ -46,18 +51,17 @@ RUN pip install --no-cache-dir .
...
-EXPOSE 8080
+# Copy over the prestart script that handles database setup.
+COPY scripts/prestart.sh /app/prestart.sh
 
-ENTRYPOINT ["gafaelfawr", "run", "--port", "8080"]
+# We use a module name other than app, so tell the base image that.  This
+# does not copy the app into /app as is recommended by the base Docker
+# image documentation and instead relies on the module search path as
+# modified by the virtualenv.
+ENV MODULE_NAME=gafaelfawr.main
```
jonathansick, line 67: Great :+1:

Dockerfile
```diff
@@ -11,17 +11,22 @@
 #   Installs the app into the virtual environment.
 # runtime-image
 #   - Copies the virtual environment into place.
-#   - Runs a non-root user.
-#   - Sets up the entrypoint and port.
+#   - Sets up additional supporting scripts.
+#   - Configures gunicorn.
 
-FROM python:3.8-slim-buster as base-image
+FROM tiangolo/uvicorn-gunicorn:python3.8-slim as base-image
```
jonathansick, line 17: Nice, I agree there's no need to use the extra fastapi docker layer.

src/gafaelfawr/cli.py
```diff
@@ -84,3 +76,7 @@ def init(settings: str) -> None:
     config_dependency.set_settings_path(settings)
     config = config_dependency()
     initialize_database(config)
+
+
+if __name__ == "__main__":
```
jonathansick, line 81: Is this a hold-over from the previous iteration, or is there something that still needs to directly execute this module?

src/gafaelfawr/database.py
```diff
@@ -29,10 +32,26 @@ def initialize_database(config: Config) -> None:
     config : `gafaelfawr.config.Config`
         The Gafaelfawr configuration.
     """
-    engine = create_engine(config.database_url)
-    initialize_schema(engine)
+    logger = structlog.get_logger(config.safir.logger_name)
+
+    # Check connectivity to the database and retry if needed.  This uses a
```
jonathansick, line 37: Nice, I like this approach

src/gafaelfawr/database.py
```diff
@@ -29,10 +32,26 @@ def initialize_database(config: Config) -> None:
...
+        try:
+            engine = create_engine(config.database_url, pool_pre_ping=True)
+            initialize_schema(engine)
+        except OperationalError:
+            logger.info("database not ready, waiting two seconds")
+            time.sleep(2)
+            continue
+        logger.info("initialized database schema")
+        break
+
```
jonathansick, line 50: Is it worth logging an explicit error if the database connection never worked?

jonathansick (approved):
Great stuff. Thanks for making those adjustments. There's just one outstanding question about the `__main__` in the `cli` module. Other than that, good to merge.

src/gafaelfawr/cli.py
rra: Yeah, it's directly executed in `prestart.sh` to initialize the database (although that will eventually be replaced with Alembic) and it's directly executed in `tox -e run` to start a development server. It's also there to support `gafaelfawr generate-key`.

src/gafaelfawr/database.py
rra: Oh, good idea, will add.

src/gafaelfawr/cli.py
jonathansick: I thought the `gafaelfawr` command-line tool was created by the entry_point in setup.cfg? https://github.com/lsst-sqre/gafaelfawr/blob/d68134dce5c5eb1036232b201b7be0f767633892/setup.cfg#L37-L39

rra: Oh, right, yes, this was only because of the workaround I was using for `prestart.sh` that isn't needed. Thanks, good catch, I forgot about that. Fixed in https://github.com/lsst-sqre/gafaelfawr/pull/131.

PR #143, opened 2021-01-15
[DM-28114] Add support for bootstrap and service tokens

rra:
- Add bootstrap tokens that have implicit admin rights
- Add the `admin:token` scope and gate admin functionality with it
- Add the remaining admins manipulation routes
- Add a new token type, service, which is created by administrators
- Add a route to create service and user tokens

The documentation is a bit incomplete since it doesn't include Helm chart and secret changes. Those will happen as part of the Gafaelfawr 2.0.0 release.


Makefile
```diff
@@ -22,3 +22,9 @@ update: update-deps init
 ui:
 	cd ui && npm run lint:fix
 	cd ui && node_modules/.bin/gatsby build --prefix-paths
+
+# Filter out errors from modules that do not have Sphinx-compatible
+# documentation.
+.PHONY: docs
+docs:
+	tox -e docs | egrep -v ' (fastapi|httpx|pydantic|starlette)\.'
```
jonathansick, line 30: So this is because you're using single-backtick syntax with API names that don't use intersphinx? I guess in principle you could use double backticks to show that you're not trying to make a link to those APIs.

jonathansick (commented):
When I tried this locally I couldn't see or set any scopes? Is this a regression?

Screen Shot 2021-01-20 at 1 16 56 PM

Screen Shot 2021-01-20 at 1 16 43 PM

rra:
To see scopes, you'll need to change `examples/gafaelfawr-dev.yaml` to add a ``group_mapping`` configuration that maps your GitHub teams to some scopes. Unfortunately there's no good way to do this without requiring local configuration changes since we can't know what groups a person running the service will be in.

Makefile
rra: I was sort of hopeful that maybe Sphinx inventories would show up in the future, and the output looks a bit better than literal text, I think, although I don't feel strongly about it.

jonathansick (approved):

PR #154, opened 2021-02-05
[DM-28666] Change /auth/analyze output to be backward compatible

rra:
Undo a poor choice I made during the FastAPI rewrite to change the
output format of /auth/analyze.  Nublado JupyterHub is using this
for now to retrieve token metadata, so convert the new token data
format to the old format we were previously generating when all
tokens were JWTs and use the same JSON structure.

This code (and route) can be dropped once JupyterHub can switch to
the new API and there's a new UI to show users their login metadata.

Also handle OpenID Connect providers that return invalid isMemberOf
structures, and ignore groups without GIDs (which are sometimes
returned by NCSA).

### DIFF
--- a/src/gafaelfawr/handlers/analyze.py
+++ b/src/gafaelfawr/handlers/analyze.py
@@ -35,24 +35,47 @@ class FormattedJSONResponse(JSONResponse):
         ).encode()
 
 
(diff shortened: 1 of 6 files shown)


cbanek (approved):
Looks good to me!

PR #176, opened 2021-03-09
[DM-29123] Improve CI caching and Docker builds

rra:
- Use the official pre-commit action with built-in caching
- Cache the entire `node_modules` directory and skip `npm ci` if possible
- Add a `.dockerignore` to reduce the amount we copy into the Docker build
- Relax some action dependencies
- Add the Python version to the tox cache key and use `tox.ini` instead of `pyproject.toml`
- Only run Docker builds for tags and ticket branches to reduce spurious image pushes
- Only run docs builds for main and ticket branches with changes touching the docs


.github/workflows/docker.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+      - name: Define the Docker tag
+        id: vars
+        run: echo ::set-output name=tag::$(echo ${GITHUB_REF} | sed -E 's,refs/(heads|tags)/,,' | sed -E 's,/,-,g')
+
+      - name: Print the tag
+        id: print
+        run: echo ${{steps.vars.outputs.tag}}
+
+      - name: Log into Docker Hub
```
jonathansick, line 25: Since you're on a modernization kick here, I'll point out that Docker itself has some GitHub Actions for things like logging in, setting up the builder, and build/pushing. That could be worth looking at if you've got some time. https://docs.docker.com/ci-cd/github-actions/

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,38 @@
+name: Docs
+
+"on":
+  push:
+    branches:
```
jonathansick, line 5: You might also want to build docs on tagged releases?

jonathansick (approved):
Great imrpovements. That pre-commit action is great.

jonathansick:
Oh, I just thought of a downside to putting the Docker build in a separate workflow rather than including it in the test workflow: it's possible to publish a docker image that fails tests. Right?

rra:
Right, I have that as a stand-up topic. It's an unfortunate trade-off.

jonathansick:
Hm, what about using [jobs.<job_id>.if](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif) to gate the Docker build rather than a separate workflow using `on`?

rra:
I looked at that but had no idea what condition to write.

PR #182, opened 2021-03-15
[DM-28898] Add Kubernetes secret management support

rra:
- Add a framework for command-line invocation of background tasks
- Add support for maintaining Kubernetes secrets containing service tokens

### DIFF
--- a/docs/_static/architecture.py
+++ b/docs/_static/architecture.py
@@ -39,6 +39,7 @@ with Diagram(
             server = KubernetesEngine("Server")
             redis = KubernetesEngine("Redis")
             storage = PersistentDisk("Redis Storage")
+            KubernetesEngine("Background")
 
             user >> ingress >> server >> redis >> storage
 
--- a/docs/applications.rst
+++ b/docs/applications.rst
@@ -183,6 +183,34 @@ Verifying tokens
 
 Tokens may be verified and used to obtain information about a user by presenting them in an ``Authorization`` header with type ``bearer`` to either of the ``/auth/v1/api/token-info`` or ``/auth/v1/api/user-info`` routes.
 
+.. _kubernetes-service-secrets:
+
+Service tokens in Kubernetes
+============================
+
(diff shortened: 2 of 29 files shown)


cbanek:
This all looks great and I am approving.  But I've got some questions as to the future direction and how this might be used for refreshing the user tokens since they won't be in the static list?  The static list in gafaelfawr makes it easy to manage since it's just static, but I wonder if maybe this isn't the place for a kubernetes CRD?  That way mobu can use the CRD to manage the token creation and the namespace (for example, if there's two mobus running or something, or the namespace is different).  Then the mobu chart can know that it is using a gafaelfawr token and the name of it, right now there's a kind of linkage there that if missing might be confusing.

I was thinking for the user tokens that putting the label on the secret would make gafaelfawr notice it, but I'm not sure that will happen with the way it is now, by design I think (so that we don't make it easy to mint tokens).

I think even if CRD was the interface it would just be a different way of triggering the same machinery.

Just a thought.  Code looks great.

cbanek (approved):

cbanek:
(Also, honestly I don't know what it would take in terms of time to make a CRD interface, so that's also an open question)

rra:
This is a great idea that didn't occur to me. I think a CRD is the right way to implement this.

It does look a fair bit more complex, and I don't think it's required for the initial Gafaelfawr 2.0 rollout (which I want to get done), so I'm going to defer this for later work. I created DM-29365 for this.

For notebook token renewal, I don't think a CRD is as clean of a fit, since we don't want to allow JupyterHub to forge the credentials of arbitrary users. It should only be able to get notebook tokens for users as they log in. The plan there is to have the hub store the token in a secret with an appropriate label and then to have Gafaelfawr refresh the token when it has lost half its lifetime, as long as it's still valid. Apologies, I should have given you a design to review first; my plan is spelled out here:

https://sqr-049.lsst.io/#kubernetes-secrets

PR #189, opened 2021-03-23
[DM-29085] Update Helm chart documentation for refactor

rra:
The Helm chart was significantly reworked to follow best practices,
which included renaming a lot of parameters.  Update the docs to
match.


rra:
Eventually I think I want to move this documentation into the charts repository and use the standard `README.md` layout used by other Helm charts, but that's a larger bit of work. This at least keeps things accurate in the meantime.

cbanek (approved):

PR #210, opened 2021-04-14
[DM-28120] Overhaul configuration documentation

rra:
We only support installing Gafaelfawr with the Helm chart, but
were documenting all configuration settings two times, sometimes
three, by trying to document the internal configuration file as
another interface.  Stop attempting to do this and only document
the Helm chart to reduce our documentation surface.

Update the Helm chart documentation for the new facilities added
for Gafaelfawr 2.0.


rra:
This probably doesn't need a super-detailed review (and it's probably rather irritating to review) because it's mostly moving things around and consolidating things. I think the main thing to review is the conceptual decision to only document the Helm chart.

cbanek:
I think the Helm chart is probably the best docs for the configuration, since it will encompass other things about the configuration at the kubernetes layer that will affect how gafaelfawr works

cbanek (approved):

PR #238, opened 2021-05-14
[DM-29365] Create Gafaelfawr service token secrets via custom objects

rra:
This PR replaces the way Gafaelfawr manages Kubernetes secrets containing service tokens.

In the first implementation, Gafaelfawr itself was configured with the list of secrets to maintain. It located and tracked those secrets with custom labels, and the Helm chart used a `CronJob` and a post-install hook to generate the secrets. The problem with this approach is that it split service configuration across charts, since Gafaelfawr needed an appropriate configuration to create a service token for a different service. It also created inobvious ordering constraints since the Kubernetes secrets were not generated continuously.

This PR abandons that approach in favor of implementing a custom resource controller. Gafaelfawr watches for changes to `GafaelfawrServiceToken` custom objects and creates `Secret` objects based on them, using a schema and approach very similar to vault-secrets-operator. On startup, it does a complete reconciliation of all custom objects, and then subsequently maintains a watch on its custom resource and reacts to change events.

The custom resource definition will be created by the Gafaelfawr Helm chart, updated in another PR attached to this same ticket.

There are several subtleties here that make this more complex than it may first appear:

- Kubernetes watches can only watch one custom resource at a time. While we currently have a single custom resource, the intent is to add more in the future (for notebook tokens and for custom ingresses), which requires maintaining multiple watches. Gafaelfawr therefore spawns a background thread to do the watch and communicates with it via a thread-safe queue.
- Correct behavior for a custom resource controller is to update the `status` component of the custom resource with the results (success or failure) of creation of the child resource. However, updating the `status` component itself triggers a `MODIFIED` event. If the `Secret` creation was successful, this is wasteful but not concerning since Gafaelfawr is idempotent: it will check the `Secret`, realize that it doesn't need changes, and do nothing. However, in the case of failure, this could create an infinite loop. Gafaelfawr therefore remembers the generation number of its custom resource and only attempts one update per generation except during the global reconcile during startup. This prevents the infinite loop at the cost of needing to touch the custom resource to force recreation of the `Secret` if the failure was unrelated to the contents of the custom resource.
- Resource metadata changes do not update the generation, but we want to synchronize labels and annotations from the parent `GafaelfawrServiceToken` to the `Secret`. Therefore, even if the generation number hasn't changed, Gafaelfawr retrieves the `Secret` and checks whether the labels and annotations are different from the `GafaelfawrServiceToken` and, if so, updates them.

Additional related changes in this PR:
- Increase the outbound HTTP timeout to 20s.
- Updated dependencies, including Sphinx, which may result in some changes to the documentation.
- When run outside of a Kubernetes cluster, Gafaelfawr uses the local Kubernetes configuration, which allows for easier interactive testing during development.
- `update-service-tokens` commands now take a `--settings` flag to make local testing easier.
- `init` handles the default for `--settings` in a different way that allows easier testing.
- Tell `pytest` that everything under `tests/*` is a test module so that it will correctly expand `assert` comparisons in helper libraries.
- Reworked the mock Kubernetes API to avoid global state.
- Improved the API documentation build and suppressed more useless warnings.


cbanek (approved):
Looks great.  Really interesting about the events / generation / metadata changing thing.

PR #250, opened 2021-05-28
[DM-30369] Fix some bugs uncovered by the recent deployment

rra:
- Display token key and type in the token change history display
- Initialize the database as part of container startup
- Add additional startup logging at the DEBUG level
- Improve error reporting if the database connection fails


rra:
This one is probably best reviewed commit by commit, since it's a bit of a grab bag. The only large change is the UI one, and nearly all of it is changes to the Selenium tests.

jonathansick (approved):
I don't think I have anything I can add — looks great :+1:

PR #265, opened 2021-06-11
[DM-30706] Display expired tokens as expired

rra:
In the UI, display expiration times of expired tokens as "expired"
instead of the delta from the current time.

Sort token lists by lats used (not yet populated), created, and
then alphabetically by token, since this will be more useful.

Reset the database before starting a Selenium test for more
reliable testing with persistent PostgreSQL.

Add an expired session token to the Selenium tests and add a test
for the new expires display behavior.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,8 @@ Change log
 3.0.2 (unreleased)
 ==================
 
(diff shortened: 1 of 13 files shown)


jonathansick (approved):
Looks good :+1:

PR #272, opened 2021-06-24
[DM-28130] Switch to Safir 2.x

rra:
Safir 2.x now provides a framework for FastAPI applications instead
of aiohttp applications.  Remove the middleware and dependencies
that are provided by it.  Use a single httpx.AsyncClient for all
requests and shut it down on application shutdown.  Use the Safir
model for the metadata returned by the internal / route.

Update dependencies and drop flake8 from the dev dependencies since
it is installed by pre-commit.


jonathansick (approved):
Looks good :+1:

PR #278, opened 2021-07-01
[DM-30971] Support pagination of GitHub team results

rra:
GitHub paginates the list of teams a user is a member of at 30
teams, and some people have considerably more teams than that.
Add support for pagination and retrieve the full list of teams.
Log all of the information we got from GitHub at debug logging
level.

### DIFF
--- a/src/gafaelfawr/models/link.py
+++ b/src/gafaelfawr/models/link.py
@@ -1,0 +1,46 @@
+"""Representation for a ``Link`` HTTP header."""
+
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Optional
+
+__all__ = ["LinkData"]
+
+_LINK_REGEX = r' *<(?P<target>[^>]+)>; rel="(?P<type>[^"]+)"'
+"""Matches a component of a valid ``Link`` header."""
+
+
+@dataclass
+class LinkData:
+    """Holds the data returned in an RFC 8288 ``Link`` header."""
+
+    prev_url: Optional[str]
+    """The URL of the previous page, or `None` for the first page."""
+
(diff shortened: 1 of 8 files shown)


src/gafaelfawr/models/link.py
```diff
@@ -0,0 +1,41 @@
...
+    """The URL of the first page."""
+
+    @classmethod
+    def from_header(cls, header: Optional[str]) -> LinkData:
+        """Parse an RFC 8288 ``Link`` with pagination URLs."""
+        links = {}
+        if header:
+            elements = header.split(",")
+            for element in elements:
+                match = re.match(' *<([^>]+)>; rel="([^"]+)"', element)
```
athornton, line 33: I see that you're basically building a hash with "prev", "next", and "first" and their targets, but it took a little doing.  Maybe name your match groups with something like "type" and "target" ?

src/gafaelfawr/providers/github.py
```diff
@@ -300,3 +304,55 @@ async def _get_user_info(self, token: str) -> GitHubUserInfo:
...
+            headers={"Authorization": f"token {token}"},
+        )
+        r.raise_for_status()
+        teams_data = r.json()
+
+        # If the data was paginated, there will be a Link header with a next
+        # URL.  Retrieve each page until we run out of Link headers.
+        link_data = LinkData.from_header(r.headers.get("Link"))
+        while link_data.next_url:
+            if not link_data.next_url.startswith(self._TEAMS_URL):
```
athornton, line 341: Does this ever happen?  Or is this just very defensive programming?

rra: I don't want us to send our GitHub token off to some random other web site if someone manages to MITM or alter the Link header or something.

athornton (approved):
Generally looks fine.  A couple of inline comments, but this seems like it should do the job.

PR #279, opened 2021-07-02
[DM-27980] Revoke GitHub OAuth grant on logout

rra:
During the first authentication from GitHub, the user chooses what
data to release to our application.  We have had persistent problems
where the user doesn't release the right data or they are not a
member of the correct organization when they first log on, and then
logging out and back in doesn't necessarily update what information
we can access via the GitHub access token.

On explicit logout, if the user logged on with GitHub, revoke the
OAuth authorization grant so that they're forced to go through the
permission screen again when they log in.  This should hopefully
avoid this issue at the cost of an extra screen when they return
after an explicit logout.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,7 @@ Change log
 - Depend on Safir 2.x and drop remaining aiohttp dependency paths.
   Remove code that is now supplied by Safir.
   Share one ``httpx.AsyncClient`` across all requests and close it when the application is shut down.
+- Fix sorting of tokens retrieved from the admin API to sort by created date before token string.
 
 3.0.3 (2021-06-17)
 ==================
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,10 +2,12 @@
 Change log
 ##########
 
-3.0.4 (unreleased)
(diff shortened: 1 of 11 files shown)


jonathansick (approved):
This all looks good code-wise. I like that this isn't a user interface change.

Is there any way that we avoid creating an OAuth authorization in the first place if the person isn't in the right org/team or is granting the appropriate information? Right now we'd have to either put up an FAQ or process a support ticket to tell the person to logout. Ideally we'd want to prevent them from thinking they're logged in to begin with (? or do I have that wrong?)

rra:
Unfortunately, no, we have to get the authorization before we have any idea what teams a user is in, and there is no way to tell GitHub to only allow people in particular teams. What we can do is detect that the user is not in any teams that we recognize and return an error page (and immediately revoke the authorization). That will be a separate story, though. (And I'll want to ask you how you think I should create the error page. Templated HTML seems bad given that we have JavaScript and will have a JS component library, but then I have to figure out how to get the error message with the valid team details into the page.)

jonathansick:
I'm not worried about maintaining some stuff in Jinja — rubin-style-dictionary still works great at that level.

PR #282, opened 2021-07-09
[DM-31056] Do not send full name in HTTP headers

rra:
HTTP headers are not guaranteed to support character sets other
than ASCII, and Starlette forces them to ISO 8859-1.  This
interferes with correctly passing the user's full name to
protected services via HTTP headers.  Stop supporting sending
the full name in X-Auth-Request-Name and require the client get
the name from the user-info endpoint if needed.

Refactor the GitHub login test suite to remove code duplication.


rra:
I think this is the proper long-term fix and is now ready for full review.

jonathansick (approved):
I agree with this change in practice. 👍

PR #286, opened 2021-07-12
[DM-31018] Return HTML errors on failure to log in

rra:
If the user failed to log in, show a human-readable error message
using an HTML template instead of returning a JSON message.  Use
a status code of 403 for all failures to log in except ones with
purely invalid URL syntax.

Additional HTML can be added to the error page by setting the
error_footer configuration option.

The HTML is currently unstyled.  Better styling will be added as
part of future UI work.


rra:
Here's how this looks when the user doesn't have any valid groups:
image

rra:
Yeah, that's a good idea. I'll leave it like this for now to get something out the door, but adding a Markdown renderer and removing the `<p>` wrapping is a great idea for the future.

PR #298, opened 2021-08-12
[DM-31386] Port to aioredis 2.x and clean up async

rra:
- Port to aioreds 2.x. This unfortunately breaks mockaioredis, so only test with real Redis and PostgreSQL for now.
- Update to Safir 2.1.0 and remove some now-duplicate code.
- Make all dependencies async so that FastAPI won't try to run them in a separate thread pool. This required some cascading changes to mark more internal functions as async and some workarounds for tests that cannot be run async due to loop issues in pytest.


src/gafaelfawr/cli.py
```diff
@@ -93,11 +93,12 @@ def generate_token() -> None:
     default=None,
     help="Application settings file.",
 )
-def init(settings: Optional[str]) -> None:
+@coroutine
```
jonathansick, line 96: This decorator is pretty cool! It is something else to port to Safir?

jonathansick (approved):
Good stuff. I'm trusting your test coverage on all the `async/await` updates :)

src/gafaelfawr/cli.py
rra: Yeah, if you have another package that wants to use it, I'd be happy to move it to Safir. I wasn't sure how common the use case of a CLI would be.

PR #299, opened 2021-08-13
[DM-31387] Handle exceptions in Kubernetes background thread

rra:
- Retry when we get a Kubernetes API exception, since this is expected in some watch cases
- Terminate the main thread when we get an unknown exception in the background watcher thread


src/gafaelfawr/storage/kubernetes.py
```diff
@@ -477,16 +479,29 @@ def run(self) -> None:
...
+                    self._api.list_cluster_custom_object,
+                    "gafaelfawr.lsst.io",
+                    "v1alpha1",
+                    self._plural,
+                )
+                for raw_event in stream:
+                    event = self._parse_raw_event(raw_event)
+                    if event:
+                        self._queue.put(event)
+            except ApiException as e:
```
athornton, line 493: Do you want to put some sort of counter on this so it's fatal after n consecutive repetitions?  Otherwise, can't we just quietly be stuck in API-is-hosed-land forever without K8s knowing there's anything wrong?

athornton (approved):
LGTM although I am slightly concerned that APIExceptions can just be retried forever.

src/gafaelfawr/storage/kubernetes.py
rra: Good call. I added a counter that resets if we successfully get an event.

PR #322, opened 2021-09-21
Auth logging info level for broken tokens.

cbanek:


cbanek:
Don't merge, just using for building

PR #347, opened 2021-11-09
[DM-32520] Update for new httpx and Docker templates

rra:
- Update for new httpx redirect handling behavior
- Switch to respx for httpx mocking
- Stop using the FastAPI Docker image and use Python 3.9.8 on bullseye
- Drop support for Python 3.8
- Incorporate dependency updates from neophile
- Update the pinned version of black
- Avoid deprecated Selenium functions


jonathansick (approved):
Looks good to me. 👍

PR #374, opened 2021-12-09
Add LDAP Group Lookup for use with OIDC Provider

brianv0:
When using OIDC Providers, add an optional LDAP group lookup to support providers that do not include groups in the ID Token claims.

This PR introduces bonsai, an LDAP library with asyncio support, and the dependencies required to run it.


brianv0:
I have tested the LDAP lookups independently at SLAC, but I have not been able to test running Gafaelfawr yet.

rra:
Merged in #390. Thank you!

PR #375, opened 2021-12-10
[DM-32899] Fix database schema initialization

rra:
Add missing await to gafaelfawr init, which was causing the database
schema to not be initialized.  Add a test for this.


frossie (approved):
lgtm

PR #391, opened 2022-01-07
[DM-33153] Convert to kubernetes_asyncio

rra:
Use the asyncio Kubernetes library, and propagate the async
status through the storage and service layer.  Convert the watch
queue to asyncio.Queue accordingly.

This allows us to throw regular exceptions from the watch task,
since it's now a normal asyncio task, and gets rid of the custom
thread pool stuff we had to do to background the watch thread.


jonathansick (approved):
(Note I struggle with PRs like this to provide value as a reviewer since there's so much going on. Maybe separating the conversation to kubernetes_asyncio for other codebase clean-ups in separate PRs would help. Essentially this review is a rubber stamp.)

rra:
This one wasn't actually ready for review; it was marked as draft. But thank you for the review! The only additional change that I'll be making is to use the new Safir support.

The changes to use kubernetes_asyncio were in a separate commit from the large cleanups. Is that enough to allow review of the interesting bit and then skipping over the commits that are just mechanical, or would it be better to have entirely separate PRs?

jonathansick:
Ah, I misunderstood because I got assigned this ticket to review in Jira.

rra:
Yeah, sorry, I should have clarified more in Jira that there were several stages of review for this. I'm never sure how to represent that.

It's fine to ask for multiple PRs! Honestly, I probably could have just self-merged the mechanical changes since there's no point in reviewing them. I'll do that in the future. Thank you for the feedback!

PR #410, opened 2022-02-08
Use LDAP query to determine UID Number of User

yee379:
This PR introduces a new configuration variable `uid_number_attr` that defines the ldap attribute name to use to lookup the user's uid number from a configured ldap directory (and when oidc is used as the authn source of the user name claim).

This PR changes `verify_oidc_token()` and `_build_token()` to allow it to ignore the uid number.

It also changes the regex for the group names `GROUPNAME_REGEX` to be more lax in allowing non-github groups to enable ldap group mappings.


src/gafaelfawr/constants.py
```diff
@@ -27,7 +27,7 @@
 CURSOR_REGEX = "^p?[0-9]+_[0-9]+$"
 """Regex matching a valid cursor."""
 
-GROUPNAME_REGEX = "^[a-z_][a-zA-Z0-9._-]*$"
+GROUPNAME_REGEX = "^[a-zA-Z0-9._-]*$"
```
rra, line 30: Do you have names that start with digits or capital letters that you have to use, or can we just filter them out instead? I ask because groups starting with digits are potentially going to cause a bunch of problems in downstream systems.

src/gafaelfawr/providers/oidc.py
```diff
@@ -199,6 +211,51 @@ async def logout(self, session: State) -> None:
...
+
+        Raises
+        ------
+        gafaelfawr.exceptions.LDAPException
+            The lookup using the uid_number_attr against the LDAP server was not
+            valid (attribute not in LDAP or resultant value not an integer).
+        """
+        assert self._ldap_config.uid_number_attr
+        attr = self._ldap_config.uid_number_attr
+        search = f"(&(Uid={uid}))"
```
rra, line 235: Does this still work if you use `uid` (with a lowercase letter)? I think that's the canonical RFC name of the attribute.

src/gafaelfawr/providers/oidc.py
```diff
@@ -199,6 +211,51 @@ async def logout(self, session: State) -> None:
...
+
+        self._logger.debug(
+            f"querying ldap {self._ldap_config.url} at {self._ldap_config.base_dn} for uid number with {search}"
```
rra, line 238: flake8 will insist that you break this into multiple strings with concatenation to keep the length down.

src/gafaelfawr/providers/oidc.py
```diff
@@ -199,6 +211,51 @@ async def logout(self, session: State) -> None:
...
+        )
+        client = bonsai.LDAPClient(self._ldap_config.url)
+        async with client.connect(is_async=True) as conn:
+            results = await conn.search(
+                self._ldap_config.base_dn,
+                bonsai.LDAPSearchScope.ONE,
+                search,
+                attrlist=[f"{attr}"],
```
rra, line 246:
```suggestion
                attrlist=[attr],
```

src/gafaelfawr/providers/oidc.py
```diff
@@ -199,6 +211,51 @@ async def logout(self, session: State) -> None:
...
+            )
+            # parse uidNumber
+            uidNumber = None
+            for result in results:
+                try:
+                    self._logger.debug(f"ldap result: {result}")
+                    uidNumber = int(result[attr][0])
```
rra, line 253: This currently implements getting the last number, but you're using a search of `ONE`, so you should only get one result. Therefore, I think you can just say:
```suggestion
                    return int(result[attr][0])
```

src/gafaelfawr/providers/oidc.py
```diff
@@ -223,6 +280,9 @@ async def _get_ldap_groups(self, uid: str) -> List[TokenGroup]:
         member_attr = self._ldap_config.group_member
         ldap_query = f"(&(objectClass={group_class})({member_attr}={uid}))"
 
+        self._logger.debug(
+            f"querying ldap {self._ldap_config.url} at {self._ldap_config.base_dn} for groups with {ldap_query}"
```
rra, line 284: Same note here on long lines.

src/gafaelfawr/verify.py
```diff
@@ -87,13 +87,17 @@ def verify_internal_token(self, token: OIDCToken) -> OIDCVerifiedToken:
...
+    async def verify_oidc_token(
+        self, token: OIDCToken, verify_uid=True
+    ) -> OIDCVerifiedToken:
         """Verifies the provided JWT from an OpenID Connect provider.
 
         Parameters
         ----------
         token : `gafaelfawr.models.oidc.OIDCToken`
             JWT to verify.
+        verify_uid : boolean
```
rra, line 99:
```suggestion
        verify_uid : `bool`
```

src/gafaelfawr/providers/oidc.py
```diff
@@ -199,6 +211,51 @@ async def logout(self, session: State) -> None:
         """
         pass
 
+    async def _get_ldap_uidNumber(self, uid: str) -> int:
```
rra, line 214: We use Python naming conventions uniformly, so:
```suggestion
    async def _get_ldap_uid_number(self, uid: str) -> int:
```

rra (commented):
Mostly just small stuff. Could you also add a test similar to the existing test for using LDAP for groups in `tests/handlers/login_oidc_test.py` (at the bottom)?

src/gafaelfawr/providers/oidc.py
yee379: fixed in bf08596ea67e12fc518bb883ae340b8c4b52a2bc

src/gafaelfawr/providers/oidc.py
yee379: good point; i was wondering if returning an error was more appropriate in case ldap returned multiple results.

src/gafaelfawr/providers/oidc.py
yee379: fixed in bf08596ea67e12fc518bb883ae340b8c4b52a2bc

src/gafaelfawr/verify.py
yee379: fixed in 4630a787a42b7e507a9eb7948cad8853bf997d3e

src/gafaelfawr/constants.py
yee379: yeah, we have some groups that starts with caps. but good point about the number tho

yee379: changed in 726ecb503e5b44c89acca4a6751fa4aa70643fe2

src/gafaelfawr/providers/oidc.py
yee379: it does. committed in https://github.com/lsst-sqre/gafaelfawr/pull/410/commits/279971917f7e2a89773798c864192a06ad718282

rra (approved):
Thanks! This looks good. Could you rebase one more time against the current `main` branch? I'll then merge it.

yee379:
updated against main. thanks @rra!

PR #419, opened 2022-02-18
[DM-33718] Refactor LDAP handling

rra:
The OpenID Connect authentication provider has acquired several bits of LDAP code to retrieve user group information and the numeric UID. Both LDAP searches were done independently, the UID attribute had to be in the same base DN as the group information (some sites maintain group membership in a separate tree from user objects), and the integration with code to pull the same information from the upstream JWT tokens was awkward.

Refactor LDAP support into a separate LDAP storage layer, and maintain a cleaner separation between the OpenID Connect authentication provider and the token verification code. This is intended to achieve the following goals:

- Provide a generalized LDAP storage layer that could potentially be used with the GitHub provider if desired
- Reuse the same LDAP connection for both UID and group lookups
- Move all token analysis into the `TokenVerifier` class and simplify the OpenID Connect provider code
- Avoid setting optional parameters in the `OIDCVerifiedToken` model in favor of explicit `get_*` functions in the verifier
- Allow the UID and groups to be in separate base DNs
- Use the specification of a UID base DN to trigger use of LDAP for UIDs, allowing the attribute to have a default value
- More consistent configuration parameter naming for parameters specifying LDAP attribute names

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,13 @@
 Change log
 ##########
 
+3.6.0 (unreleased)
+==================
+
+- Add support for retrieving the user's numeric UID from LDAP when authenticating with an OpenID Connect provider.
+- Add required dependency for LDAP support to the Docker image.
+- Update dependencies.
+
 3.5.1 (2022-01-14)
 ==================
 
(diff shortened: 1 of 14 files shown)


jonathansick (approved):
This all looks great. I'm not experienced in LDAP, but the code looks good.

PR #464, opened 2022-04-08
[DM-34334] Standardize on using files to pass in secrets

rra:
We were using environment variables for some secrets and files
for other settings.  Standardize on files for all secrets and add
some background to the architecture documentation.

### DIFF
--- a/docs/arch/configuration.rst
+++ b/docs/arch/configuration.rst
@@ -35,19 +35,28 @@ Pydantic and dataclasses
 ========================
 
 It would be ideal if Pydantic could be used directly for settings without rewriting the Pydantic ``Settings`` class into dataclasses.
-However, there are two missing features in the Pydantic system that interfere with this:
+However, there is a missing feature in the Pydantic system that interfere with this.
 
-#. Loading secrets from disk directly in the Pydantic model is difficult.
-   Pydantic does support a mechanism for loading configuration keys from disk files, but it doesn't support nested structure, which we want so that the configurations for different internal Gafaelfawr components are kept separate (which in turn simplifies a lot of code that otherwise would have to check for `None` repeatedly).
-#. Pydantic provides poor support for loading a YAML file whose file name is not known statically.
-   We have to use ``parse_obj``, which in turn makes some other Pydantic Settings constructor arguments accessible only via static configuration.
+There are two options for loading Kubernetes secrets in a pod.
+One is to inject them via environment variables.
+The other is to expose the secrets as files in a temporary file system created by Kubernetes.
+Unfortunately, neither of these work with Pydantic given the constraints of Gafaelfawr.
+
+The problem with the environment variable approach is that Pydantic does not really support setting individual keys of a nested configuration model via environment variables.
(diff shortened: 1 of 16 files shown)


rra:
This is mostly just refactoring and pretty obvious mechanical changes; feel free to review only for the concept as spelled out in the changes to the architectural documentation and not worry about the code details.

docs/arch/configuration.rst
```diff
@@ -29,3 +30,36 @@ One environment variable may be of interest for running a development copy of Ga
...
     Normally this is handled by either the ``tox -e run`` command, the ``tox -e docker`` command, or the release Docker image.
+
+Pydantic and dataclasses
+========================
+
+It would be ideal if Pydantic could be used directly for settings without rewriting the Pydantic ``Settings`` class into dataclasses.
+However, there are two missing features in the Pydantic system that interfere with this:
+
+#. Loading secrets from disk directly in the Pydantic model is difficult.
+   Pydantic does support a mechanism for loading configuration keys from disk files, but it doesn't support nested structure, which we want so that the configurations for different internal Gafaelfawr components are kept separate (which in turn simplifies a lot of code that otherwise would have to check for `None` repeatedly).
```
jonathansick, line 41: I'm trying to grasp this argument. Pydantic `parse_obj` can deal with nested models just fine, right? Oh, is it that `BaseSettings` can't be nested, whereas `BaseModel` types can have nesting? And we want `BaseSettings` because that deals with environment variables...

Or are you saying that you want a monolithic YAML config file, but then have different settings objects read out different parts of it?

I wonder if now that Gafaelfawr isn't using environment variables, if these BaseSettings types could actually be BaseModel because they're really models for the configuration YAML files? Would that help?

jonathansick (approved):
I just want to double-check, is this classmethod on `Settings` still needed? It seems to deal with merging environment variables into the config: https://github.com/lsst-sqre/gafaelfawr/blob/ce33fc03f067c4fb1121447163ebc0c9da0ed0a9/src/gafaelfawr/config.py#L294-L309

docs/arch/configuration.rst
rra: Correct, the problem is with using environment variables.

It works if I don't use environment variables, but if I don't use environment variables then all the secrets have to be in files and then I need to read those files and materialize the secrets, which is the big motivator for dynamically rewriting the configuration to cache the secrets in memory. I suppose the correct way to fix this is to add a function to retrieve a secret from a file and use it whenever the secret is needed rather than caching the secret in memory (which would also provide a mechanism for doing a smarter secret management system in the future). Maybe I should go that route.

I could switch from `BaseSettings` to `BaseModel` now, I think, since I'm not sure I'm using any of the functionality of `BaseSettings`, but I'm not sure it makes any difference? I still need to have two sets of configuration because the one read from disk has pointers to a bunch of paths for files that need to be materialized into actual secrets.

I'll try to explain this better.

rra:
> I just want to double-check, is this classmethod on `Settings` still needed? It seems to deal with merging environment variables into the config:

Yup, good catch, this is no longer needed.

PR #475, opened 2022-04-27
[DM-34335] Support UID/GID assignment using Google Firestore

rra:
Add support for configuring Firestore (using workload identity for authentication and authorization) as a repository of UID and GID assignments.  When this option is enabled, Gafaelfawr will ignore UID and GID information from the upstream OpenID Connect ID tokens or from LDAP and instead assign UIDs and GIDs on first use of that username or group name, or look up the UID and GID from Firestore if one was already assigned to that username or group name.

To implement this support, retrieval of user information was extracted from the OpenID Connect authentication provider into a new internal service, and the cache implementation previously used only for internal and notebook tokens was made more generic and also used to cache UIDs and GIDs.  UID and GID information is cached for the lifetime of the process, since UIDs and GIDs are never reused and should never change.

UID and GID assignment is done from the ranges documented in [DMTN-225](https://dmtn-225.lsst.io/), including assigning UIDs to bot users from a different range than the UIDs for regular users.

This PR also defers group information from LDAP so that it's no longer stored with the token data and instead is retrieved dynamically when needed, such as in `/auth` responses and responses to the user-info API. This allows changes to group memberships to be seen dynamically without requiring reauthentication, although the scopes of a token will not change.

Some refactoring work remains to be done, most notably:

- Further clean up the API of the caching layer and make it more generic
- Add helper function to do OpenID Connect login tests
- Improve the user information service class structure to get rid of `RuntimeError`
- Move code to determine scopes from groups to the user information service

Those will be mostly mechanical changes and thus not interesting to review, so I'll do them in a separate self-merged PR.

Some follow-up work remains, namely:

- Cache group information from LDAP for a short period of time to avoid excess queries from a quick sequence of `/auth` calls
- Support the new user information service with GitHub authentication
- Support getting the user's full name and email address from LDAP

Those will be handled in subsequent PRs, probably after the 4.1.0 release.

This PR also refreshes the JavaScript dependency information to reflect recent dependabot updates and uses `--legacy-peer-deps` to work around the fact that react-aira-modal seems to be dead upstream (we need to eventually do something about that).


rra:
Note that DMTN-225 is currently slightly out of date (it doesn't have a bot UID range and therefore the other ranges are a bit off). I'm fixing that in another PR.

src/gafaelfawr/services/userinfo.py
```diff
@@ -0,0 +1,470 @@
...
+    In some cases, we take user metadata from external systems.  Examples are:
+
+    #. Resolve a unique identifier to a username via LDAP.
+    #. Get user group membership from LDAP.
+    #. Get UID or GID from LDAP.
+    #. Assign and manage UIDs and GIDs via Google Firestore.
+
+    This service manages those interactions.  UID/GID data from Firestore is
+    cached.  LDAP data is not cached, since LDAP is supposed to be able to
+    handle a very high query load.
```
jonathansick, line 38: Ah, so the caching is not about latency to the data but about respecting the backend service?

jonathansick (approved):
Looks good!

On the JavaScript model dependency issue, yeah; we'll have to tackle migrating the UI into Squareone... perhaps sooner than later.

The caches and their locking is very neat. I'll keep that design in mind if I need to add an in-app cache in the future.

Thinking bigger picture, what's the change for folks like me who run bot users? I know that now we need to prefix our bots' usernames with `bot-`. How should we be getting the uid? In noteburst, I do something based on mobu where I've created the uid for a given username on my side through configuration. [My "login" method](https://github.com/lsst-sqre/noteburst/blob/ffe0eb9ba20fd8865fbd96cd4d194ec498462743/src/noteburst/jupyterclient/user.py#L53) with the `POST /auth/api/v1/tokens` endpoint needs both the username and UID. What's the recommended way for apps to now get their UID?

src/gafaelfawr/services/userinfo.py
rra: This comment is probably inaccurate on further reflection, and I'm going to put in a five-minute cache for all LDAP information. I'm not worried about the latency during initial login, since that's going to be kind of slow anyway, but if the user is making a flurry of authenticated API requests, we probably need to prevent that from turning into a flurry of LDAP queries, particularly since we don't maintain a persistent connection right now. I'll do that in a separate PR.

src/gafaelfawr/models/token.py
```diff
@@ -124,7 +124,9 @@ class TokenGroup(BaseModel):
         regex=GROUPNAME_REGEX,
     )
 
-    id: int = Field(..., title="The numeric GID of the group", example=123181)
+    id: Optional[int] = Field(
```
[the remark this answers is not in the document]
rra: Yeah, I'm not really happy with this whole design, since the model as returned to the user will always have that field populated and the only reason why it's optional is because the same model is being used internally to pass data around. I need to do the refactoring work to split it into two models, and have the public model not be `Optional`. I've added that to my to-do list.

rra:
I think I can make the cache even better. I'll send you that as a separate PR when I figure out how I want to do it.

For bot users, for now you can keep doing what you have been doing and it will continue to work, but eventually I want to have Gafaelfawr automatically assign UIDs for those users when created via the API as well. The tricky part is what to do on deployments that don't use Firestore, since in that case the API client needs to still provide the UID. That's a bit hard to make clear in the API.

PR #506, opened 2022-06-09
[DM-34613] Further LDAP integration improvements

rra:
- Service tokens now must be for bot users, meaning the username must start with `bot-`
- Uniformly apply the principle that data stored with the token overrides data from external sources
- Treat UID the same as other data and retrieve it on the fly as needed rather than storing it with the token
- Minor updates to the architecture documentation (more coming in a larger overhaul of the tech notes)


jonathansick (approved):
Looks good 👍

PR #513, opened 2022-06-15
[DM-35238] Generate API documentation from OpenAPI spec

rra:
Use sphinxcontrib-redoc to generate REST API documentation at
api.html.  Move the internals Python API documentation to
arch/internals, since only developers will need to look at it.
In support of this, add a CLI command to spit out the OpenAPI
schema so that it can be used in the tox build.
 
I experimented with sphinxcontrib-openapi, but it appears to not
be maintained and has bitrotted, and I couldn't get it to work
with the OpenAPI schema generated by FastAPI.  Redoc does,
provided that I override the version of Redoc with the current
upstream release.

There is as yet no back link from the API documentation to the
rest of the documentation.

Now that the API documentation is of the REST API, move it up in
the documentation index.

Remove now-unnecessary documentation pins and fix an incorrect
bit of local Docker configuration.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -128,6 +128,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+        with:
+          # Ensure the documentation gets the right version.
+          fetch-depth: 0
 
       - name: Set up Python
         uses: actions/setup-python@v4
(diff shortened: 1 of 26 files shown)


jonathansick (approved):
Very nice, thanks for figuring this out. I suppose we could have some code that processes the OpenAPI spec to add Sphinx-context specific things like a relative backlink to the Sphinx docs.

rra:
Yeah, it looks like the easiest way to do that is to provide a custom template file, which I believe is the HTML wrapper into which the Sphinx plugin embeds the Redoc invocation.

PR #544, opened 2022-07-29
[DM-35644] Switch to new pyproject-only build system

rra:
Follow the new template for SQuaRE library packages and switch to
a pure pyproject.toml build system.  Remove setup.cfg and setup.py.

This changes how the project URLs are stored, which means the
standard Safir get_metadata function no longer works.  Inline the
new version of it.


src/gafaelfawr/handlers/index.py
```diff
@@ -25,6 +29,11 @@ async def get_index() -> Metadata:
...
-    return get_metadata(
-        package_name="gafaelfawr", application_name="gafaelfawr"
+    pkg_metadata = cast(Message, metadata("gafaelfawr"))
+    return Metadata(
+        name="gafaelfawr",
+        version=pkg_metadata["Version"],
+        description=pkg_metadata["Summary"],
+        repository_url=get_project_url(pkg_metadata, "Source"),
+        documentation_url=get_project_url(pkg_metadata, "Homepage"),
     )
```
jonathansick, line 39: Yeah, interesting question with this. I had a bit of difficulty trying to figure out what project URL was equivalent to the old homepage from setup.cfg and it wasn't clear at all.

Basically you've got [urls](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#urls) in the TOML file, and those docs say the corresponding "core metadata"field in [Project-URL](https://packaging.python.org/en/latest/specifications/core-metadata/#project-url-multiple-use) — but I don't know what maps to [Home-page](https://packaging.python.org/en/latest/specifications/core-metadata/#id15).

I guess we could either update Safir's get_metadata or just move the onnus on building the Metadata object to the user like you've done here.

jonathansick (approved):

src/gafaelfawr/handlers/index.py
rra: The term `Home-page` doesn't appear anywhere in https://packaging.python.org/en/latest/specifications/declaring-project-metadata/ but I don't know if the omission was intentional. Poetry supports a `homepage` key (https://python-poetry.org/docs/pyproject/), but setuptools appears not to. I'm not sure if this is just one of those things that hasn't been sorted out because the support is still in beta.

PR #549, opened 2022-08-04
[DM-35842] Fix editing of token scopes

rra:
When a user token was edited to change its scope, but not its
expiration time, its scopes were not updated in Redis.  Since Redis
is canonical for token scopes, this meant that the change appeared
to go through but had no actual effect.  Fixed by updating Redis if
either the scope or expiration of a user token is changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ The internal configuration format may change in minor releases.
 Dependencies are updated to the latest available version during each
 release.  Those changes are not noted here explicitly.
 
+## 5.0.3 (unreleased)
+
+- When a user token was edited to change its scope, but not its expiration
+  time, its scopes were not updated in Redis.  Since Redis is canonical
+  for token scopes, this meant that the change appeared to go through but
+  had no actual effect.  Fixed by updating Redis if either the scope or
+  expiration of a user token is changed.
+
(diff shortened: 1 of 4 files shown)


jonathansick (approved):
Are editable token scopes a project requirement? To be honest I didn't realize we could edit the scopes or expiration dates on existing tokens, though I do see it in the UI now. I don't think I've seen editable tokens on other services. Implementation looks good, though.

rra:
They're not a project requirement in the sense that they're not in any of the higher-level project requirement documents maintained outside of our group. My logic for adding the feature is that I'm expecting our users to be relatively unsophisticated users of tokens, even compared to, say, the typical GitHub token user, and once they've gotten a token pasted in everywhere, it's kind of a pain to have to change it if they discover that they need an additional scope. I couldn't think of a security reason not to allow that, so went with it.

I uncovered this bug because I had a token that I wanted to add `read:image` scope to so that I could also use it for inspecting the HiPS service and was surprised when that didn't work, so that's a bit of evidence in favor of the feature. That said, it's probably worth a discussion at some point; I'll queue it up for the next stand-up. (I'll go ahead and merge this fix for now, though, and we can have a separate discussion about whether the feature should go away at some point in the future.)

PR #554, opened 2022-08-09
[DM-28121] Add background maintenance task

rra:
Add a background maintenance task that:

- Removes database entries and adds change history entries for expired tokens
- Truncates the change history table to keep it from growing without bound

### DIFF
--- a/src/gafaelfawr/cli.py
+++ b/src/gafaelfawr/cli.py
@@ -192,7 +192,7 @@ async def maintenance(settings: Optional[str]) -> None:
         config_dependency.set_settings_path(settings)
     config = await config_dependency()
     logger = structlog.get_logger("gafaelfawr")
-    logger.debug("Starting")
+    logger.debug("Starting background maintenance")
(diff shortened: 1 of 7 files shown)


jonathansick (approved):
I hadn't seen a SQLAlchemy query with the `In_` from a previous query like that before, so it was nice to learn.

Are you you planning on running this via a Kubernetes cron job deployed as part of the Helm deployment?

rra:
Yeah, `CronJob`. It doesn't feel like it needs to run constantly, and while I could integrate it with the Kubernetes operator, it seems cleaner to let Kubernetes do process management and not try to juggle multiple threads of execution. I figure I'll run this one hourly and then run a data audit nightly to look for inconsistencies.

rra:
Oh, I could optimize the first two queries into one query, couldn't I. No need to search for the tokens and then get the other columns when I can do that all at once. That's left over from an earlier iteration when I thought I had to chase child tokens. I'll make that change tomorrow.

PR #555, opened 2022-08-09
[DM-35872] Add Slack alerts for uncaught exceptions

rra:
Use a wrapped route handler class to catch uncaught exceptions and
report them to a Slack webhook if so configured.

This change is primarily a proof-of-concept for Slack alerting,
which will be used in the future for reporting problems found by
periodic data audits and may be used to report particularly serious
errors that may indicate a broader system problem.  As of this
change, it only handles uncaught exceptions (and tests that none of
the normal errors we are already handling produce Slack messages).


src/gafaelfawr/config.py
```diff
@@ -318,6 +318,9 @@ class Settings(BaseSettings):
     error_footer: Optional[str] = None
     """HTML to add (inside ``<p>``) to login error pages."""
 
+    slack_webhook_file: Optional[str] = None
+    """File containing the Slack webhook to which to post alerts."""
```
jonathansick, line 322: Is this the file contents or the file path? Ah, its the filepath, right.

jonathansick (approved):
Very neat. I'm sure we could use this in Safir as well.

Do you think that Gafaelfawr should also emit a lot message for things it also posts to Slack? Or does it already ( maybe I missed that)? If not, maybe it'd be helpful as a way to go from seeing an issue in Slack to also then correlate it to logs for further analysis.

src/gafaelfawr/config.py
rra: Yeah, maybe I should change all those `_file` instances to `_path`? Or `_filepath`? Happy for any advice there.

rra:
I think it does issue such a message, although this isn't super-obvious. But the custom route handler raises the exception again after it reports it to Slack, so it remains an uncaught exception that is then picked up by uvicorn and will be logged following uvicorn's normal logging practice.

PR #560, opened 2022-08-17
[DM-35881] Fix primary GID assignment with Firestore

rra:
In the Firestore case, a variable was reused incorrectly, causing
the primary GID of the user to be set to the GID of their last
group instead of the user private group.  Fix and add a test case.


rra:
Wrong branch name.

PR #567, opened 2022-08-23
[DM-35959] Add periodic audit command

rra:
Add a command that does a consistency audit of the Gafaelfawr
data stores and reports any problems to Slack.  This contains a
selection of simple and obvious checks, and we can add more checks
later based on problems we've seen in practice.

### DIFF
--- a/src/gafaelfawr/factory.py
+++ b/src/gafaelfawr/factory.py
@@ -420,43 +420,6 @@ class Factory:
             # This should be caught during configuration file parsing.
             raise NotImplementedError("No authentication provider configured")
 
-    def create_user_info_service(self) -> UserInfoService:
-        """Create a user information service.
-
-        This service retrieves metadata about the user, such as their UID,
-        groups, and GIDs.  This is the generic service that acts on Gafaelfawr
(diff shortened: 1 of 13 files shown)


src/gafaelfawr/factory.py
```diff
@@ -420,43 +420,6 @@ def create_provider(self) -> Provider:
             # This should be caught during configuration file parsing.
             raise NotImplementedError("No authentication provider configured")
 
-    def create_user_info_service(self) -> UserInfoService:
```
rra, line 423: Apologies, this is just a code move. I forgot I did that or I would have merged it separately.

jonathansick (approved):
This looks great.

The only thing I was wondering about was whether you'd want the audit report to be available elsewhere than Slack, given the message length limit. For example, persisting a report in Redis or Postgres and then being able to grab that report from the REST API and integrate with admin tools to resolve consistency issues. But that's probably overkill for now.

rra:
This is a great point. I think what I'll do for the interim is add verbose logging of the audit findings with some more details, which is useful anyway and means that we can also recover the exact findings from the pod logs or Log Explorer.

PR #584, opened 2022-09-12
[DM-35879] Log uvicorn messages in JSON

rra:
- Redirect uvicorn messages through structlog to add JSON formatting
- Parse the uvicorn access messages to populate a Google-style `httpRequest` log attribute

The motivation of this set of changes is to standardize on JSON as the log format for all Gafaelfawr logs, allowing easier log queries and exploration in systems like Google Log Explorer that automatically understand JSON structure and allow individual fields to be used in queries and filters. As a secondary benefit, this will also avoid misclassification by Google of pure text logs as errors when they're intended to have a lower severity, by lifting the actual Python log severity into JSON in the field where Google expects it.

### DIFF
--- a/src/gafaelfawr/logging.py
+++ b/src/gafaelfawr/logging.py
@@ -63,7 +63,7 @@ def process_uvicorn_access_log(
 def setup_uvicorn_logging(loglevel: str = "INFO") -> None:
     """Set up logging.
 
-    This configures the uvicorn to use structlog for output formatting.  It is
+    This configures uvicorn to use structlog for output formatting.  It is
     used by the main FastAPI application.
 
     Parameters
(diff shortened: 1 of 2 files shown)


src/gafaelfawr/logging.py
```diff
@@ -0,0 +1,123 @@
...
+    event_dict["httpRequest"]["requestMethod"] = method
+    event_dict["httpRequest"]["requestUrl"] = url
+    event_dict["httpRequest"]["status"] = match.group(4)
+    return event_dict
+
+
+def setup_uvicorn_logging(loglevel: str = "INFO") -> None:
+    """Set up logging.
+
+    This configures the uvicorn to use structlog for output formatting.  It is
```
jonathansick, line 66:
```suggestion
    This configures uvicorn to use structlog for output formatting.  It is
```

jonathansick (approved):
This is brilliant. Can we get this into Safir too?

PR #592, opened 2022-09-20
DM-29094: Support a required minimum lifetime

rra:
Add a parameter to the /auth route to specify a required minimum lifetime.  This is measured against the lifetime of any delegated token that we would pass on to the service, such as a notebook token for the Notebook Aspect (the motivating application), but it can be used as a check against the user's session token lifetime if desired.

This is a bit more complicated than immediately apparent since we have to do some careful logic to extrapolate from the lifetime of the user's authentication token to the lifetime of a delegated token, and we have to support invalidating the cache and generating a new delegated token even if more than half of its lifetime is still remaining.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Dependencies are updated to the latest available version during each release. Th
 
 ### New features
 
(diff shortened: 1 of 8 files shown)


jonathansick (approved):
Looks good. I think I follow the logic from all the code comments (thanks!)

PR #593, opened 2022-09-20
DM-36307: Make delegated scopes optional

rra:
Currently, if a service requests a delegated internal token with a list of scopes specified in delegate_scope, those scopes also become mandatory for authentication to that service to succeed, just as if they were listed in scope.

This ensures behavior is consistent, but in practice it has proven too limiting.  For example, the Portal Aspect would like to have read:image and read:tap scopes so that it can access those services on behalf of the user, but it is still usable (with reduced functionality) if those scopes are not available, and we don't want to block all Portal Aspect access to a user who does not have image access for some reason.

Correct this by making scopes in delegate_scope optional.  If the authenticating token has a scope that's requested in delegate_scope, it will be included in the internal token.  But if it does not, the authentication will still succeed and the delegated token will still be created, but without that scope.  To restore the previous behavior of making each scope mandatory for authentication as well, add the scope to the scope parameter as well, and either omit satisfy or set satisfy=all.

The only service that was using delegate_scope was the Portal Aspect, and it wants the new behavior, so this should not pose a significant backward-compatibility security risk.


docs/applications.rst
```diff
@@ -141,9 +141,12 @@ The URL in the ``nginx.ingress.kubernetes.io/auth-url`` annotation accepts sever
 
 ``delegate_scope`` (optional)
     A comma-separated list of scopes that the internal token should have.
```
jonathansick, line 143:
```suggestion
    A comma-separated list of scopes that the internal token should have, if available from the authenticating token.
```

jonathansick (approved):
You might want to double check the sentence on internal tokens in https://gafaelfawr.lsst.io/arch/scopes.html#scope-inheritance

rra:
Oh, good catch, thanks.

PR #599, opened 2022-09-29
DM-36366: Rewrite documentation using new Documenteer

rra:
Rework the documentation to use the new Documenteer system and theme.  Use this as an opportunity to restructure the documentation in a way that will hopefully be more logical and oriented towards the reader, rather than when I happened to think of something.

Use sphinx-diagrams to generate the architecture diagram instead of generating it outside of Sphinx.

Create the directory where the OpenAPI schema is written if it didn't already exist.


docs/user-guide/administrators.rst
```diff
@@ -0,0 +1,24 @@
...
+
+Bootstrapping
+=============
+
+Gafaelfawr can be configured with a special token, called the bootstrap token.
+This token must be generated with ``gafaelfawr generate-token`` and then stored in the ``bootstrap-token`` key of the Gafaelfawr Vault secret.
+See :ref:`vault-secrets` for more details.
+It can then be used with API calls as a bearer token in the ``Authenticate`` header.
+
+The bootstrap token acts like the token of a service or user with the ``admin:token`` scope, but can only access specific routes, namely ``/auth/api/v1/tokens`` and those under ``/auth/api/v1/admins``.
```
jonathansick, line 24: I'll have to think of a way to set up a role for cross-referencing into the Redoc docs.

PR #604, opened 2022-10-03
DM-36447: Add support for fixing some audit issues

rra:
Add a flag --fix to the gafaelfawr audit command that attempts to fix some issues, specifically the ones that have turned up to date in the IDF.  (The theory is that I'll only bother to write fix code for problems that we've seen.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Versioning follows [semver](https://semver.org/). Versioning assumes that Gafael
 
 Dependencies are updated to the latest available version during each release. Those changes are not noted here explicitly.
 
+## 6.1.0 (unreleased)
+
+### New features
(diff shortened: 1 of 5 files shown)


jonathansick (approved):
This looks great. Only thing I'm wondering about is if we've got docs here or in Phalanx on using the audit CLI. I'm guessing we're configuring a k8s Job and by default running `--fix`? Rather than logging into a Pod's container and running the CLI? But we might want to note somewhere that this is happening.

rra:
Yeah, I will definitely add some documentation for how to run this. My short-term plan is `kubectl exec gafaelfawr-pod -- gafaelfawr audit --fix`. My longer-term plan is an admin UI and then making this action available via an authenticated REST call.

PR #611, opened 2022-10-13
DM-36605: New GitHub groupMappings, fix tox -e run

rra:
- Add the new syntax to `groupMappings` for specifying a GitHub org/team pair
- Fix `tox -e run` (broken in 4.0.0)


rra:
I'm not sure whether I'm happy with this syntax. Definitely interested if you have any other ideas.

jonathansick:
I think the full object mapping with a github key, while verbose, is definitely future proof.

Another approach could be:

```yaml
kind: "github"
organization: "lsst-sqre"
team: "square"
```

But although that might look simpler I imagine it'd mean more work on the schema side if there are more "kinds" (or use provider or whatever relevant term).

PR #618, opened 2022-10-18
DM-36458: Convert Kubernetes operator to Kopf

rra:
In preparation for adding more Kubernetes operator functionality, convert the Kubernetes operator to Kopf.  The Kopf framework now handles monitoring Kubernetes for changes, deduplicating events, and storing status in Kubernetes, which simplifies the internal logic (at the cost of having somewhat less control over how status is stored).

Unfortunately, Kopf does not support testing against a mock Kubernetes, so testing is now done only via a new tox py-full target that enables Kubernetes tests using the current default Kubernetes environment.  In CI, this is done using Minikube.

For now at least, remove the tests for specific logging from the Kubernetes operator, since integrating that with Kopf and ignoring Kopf's own logs is tricky and the value of the tests is not that high.


src/gafaelfawr/operator/startup.py
```diff
@@ -0,0 +1,43 @@
...
+from safir.kubernetes import initialize_kubernetes
+
+from ..dependencies.config import config_dependency
+from ..factory import Factory
+
+__all__ = ["startup", "shutdown"]
+
+
+@kopf.on.startup()
+async def startup(memo: kopf.Memo, **_: Any) -> None:
```
jonathansick, line 19: Oh neat. I didn't know/remember this memo data storage.

src/gafaelfawr/operator/__init__.py
```diff
@@ -0,0 +1,17 @@
+"""Kubernetes operator framework.
+
+This module imports all of the handlers for Gafaelfawr's Kubernetes operator
+and serves as an entry point for Kopf_.
+
+Examples
+--------
+Start the Gafaelfawr Kubernetes operator with the following command:
+
+.. code-block: shell
```
jonathansick, line 10: Interesting. The module docstring is getting cut off here. https://gafaelfawr.lsst.io/v/DM-36458/dev/internals.html#module-gafaelfawr.operator I didn't expect that to happen.

I guess arguably that documentation could go in the user guide for gafaelfawr rather than the docstring.

jonathansick (approved):
This looks great.

docs/dev/development.rst
```diff
@@ -94,13 +94,41 @@ This uses tox-docker to start PostgreSQL and Redis Docker containers for the tes
...
 On Debian and Ubuntu systems, you can install this with ``apt install chromium-driver``.
 
+To run the tests with coverage analysis and generate a report, run:
+
+.. code-block:: sh
+
+   tox -e py-coverage,coverage-report
+
+To test the Kubernetes operator, you must have a Kubernetes cluster available that is not already running Gafaelfawr.
+This is only tested with Minikube_, which is the approach used by CI.
```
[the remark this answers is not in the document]
rra: I've personally never gotten Minikube to work, but I'll add some documentation based on the Phalanx documentation.

src/gafaelfawr/factory.py
```diff
@@ -194,15 +235,9 @@ async def standalone(
...
-        logger = structlog.get_logger("gafaelfawr")
-        statement = select(SQLAdmin) if check_db else None
-        session = await create_async_session(engine, statement=statement)
-        try:
-            context = await ProcessContext.from_config(config)
-            async with aclosing(context):
-                yield cls(context, session, logger)
-        finally:
-            await session.remove()
+        factory = await cls.create(config, engine, check_db)
```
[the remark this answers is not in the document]
rra: Good catch, done.

src/gafaelfawr/operator/__init__.py
rra: I only had one colon, so it thought this was a comment. That said, moving this is a good idea, since no one will find it there. I added it to the CLI page in the user-guide instead.

PR #646, opened 2022-12-02
DM-37192: Invalidate the LDAP cache if login was rejected

rra:
If we reject the login because the user is not a member of any valid groups, invalidate the LDAP cache so that the next attempt will check LDAP again.  The user is probably going to try to fix the problem right away, and we don't want to make them wait for the LDAP cache to time out before they can see if their fix worked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ Dependencies are updated to the latest available version during each release. Th
 - All commands that took a `--settings` option to specify the path to the configuration file now take a `--config-path` option instead. This name is clearer and avoids introducing a separate "settings" term.
 - The default path to the Gafaelfawr configuration file is now taken from the `GAFAELFAWR_CONFIG_PATH` environment variable rather than `GAFAELFAWR_SETTINGS_PATH`, for the same reason.
 
+### Bug fixes
+
+- If a user's login was rejected because they were not a member of any known groups, invalidate the LDAP cache for that user before returning the error. The user is likely to immediately try to fix this problem, and making them wait until the LDAP cache times out to see if the fix worked is confusing.
+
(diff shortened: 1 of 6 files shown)


jonathansick (approved):
Good idea. Now, is Gafaelfawr only run as a `replicaCount: 1` Deployment? If there are multiple replicas I wonder if this would introduce some inconsistency in behaviour — sometimes the LDAP fix would take right away, and sometimes not.

rra:
I worried about that, but I think it's fine. Here's my logic (and I'll include this in a docstring or something): Retrieving LDAP information for a user only happens either (a) during the login process, or (b) when checking a user's token (via the `/auth` route, the API, etc.). For the typical case of a user who hasn't onboarded yet, (b) is impossible since they haven't previously authenticated and have no tokens. When they go through the login process, we will retrieve their LDAP information and cache it, but then we ask whether they're authorized. If not, we don't issue them a token, and with this diff, we then invalidate the cache. These both happen as part of processing the same request, so they can't be split across multiple instances.

Therefore, in that normal case, I believe the effect of this is to remove the information that we ourselves just cached, and no other instance of Gafaelfawr could have cached LDAP data because the user has not successfully authenticated.

Where this problem does come up is if the user had previously authenticated, and then their information changed such that they're no longer a member of an eligible group. In that case, the cache invalidation done by the login process may be undone by the user using an existing unexpired token to authenticate to something else. But I think this case is relatively rare; in the case where we are intentionally attempting to kick the user off the system, we should delete all of their tokens (which reminds me to add to the to-do list some API to do that), which would prevent this case.

So there is some chance of confusing behavior if the user's group membership was working and then broke, but hopefully this will be rare and this diff should safely handle the initial user case.

PR #655, opened 2022-12-14
DM-37291: Support passing delegated token in Authorization

rra:
Add new config.delegate.useAuthorization GafaelfawrIngress setting and use_authorization query parameter to the /auth route that also puts a copy of any delegated token into the Authorization header as a bearer token.  The CADC TAP service requires the token be passed that way to work well with their authentication layer, and this removes the need for a custom NGINX configuration snippet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ Dependencies are updated to the latest available version during each release. Th
 ### New features
 
 - The response from the `/auth` now reflects `Authorization` and `Cookie` headers from the incoming request with Gafaelfawr tokens and secrets filtered out. `GafaelfawrIngress` resources use this to filter those secrets out of the request passed to the protected service, avoiding leaking user credentials to services. Manual ingress configurations should add `Authorization` and `Cookie` to the `nginx.ingress.kubernetes.io/auth-response-headers` annotation to get the benefits of this filtering.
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Dependencies are updated to the latest available version during each release. Th
 ### New features
 
 - The response from the `/auth` now reflects `Authorization` and `Cookie` headers from the incoming request with Gafaelfawr tokens and secrets filtered out. `GafaelfawrIngress` resources use this to filter those secrets out of the request passed to the protected service, avoiding leaking user credentials to services. Manual ingress configurations should add `Authorization` and `Cookie` to the `nginx.ingress.kubernetes.io/auth-response-headers` annotation to get the benefits of this filtering.
+- Added a `config.delegate.useAuthorization` field in `GafaelfawrIngress` and a `use_authoriation` query parameter for the `/auth` route that, if set, also puts any delegated token in the `Authorization` header, as a bearer token, in the request sent to the protected service. This allows easier integration with some software that expects tokens in standard headers rather than Gafaelfawr's custom `X-Auth-Request-Token` header.
```
jonathansick, line 17:
```suggestion
- Added a `config.delegate.useAuthorization` field in `GafaelfawrIngress` and a `use_authorization` query parameter for the `/auth` route that, if set, also puts any delegated token in the `Authorization` header, as a bearer token, in the request sent to the protected service. This allows easier integration with some software that expects tokens in standard headers rather than Gafaelfawr's custom `X-Auth-Request-Token` header.
```

jonathansick (approved):

PR #657, opened 2022-12-15
DM-37291: Add support for anonymous ingresses

rra:
Allow a GafaelfawrIngress to specify that it is anonymous.  In this case, most of the configuration is not allowed, but Gafaelfawr is still configured as the auth-uri so that it can strip tokens and cookies from the Authorization and Cookie headers.  Add a new route, /auth/anonymous, that does only this stripping.

Also diagnose a GafaelfawrIngress that requests a basic authentication challenge but also wants an auth redirect, since those are mutually incompatible.

### DIFF
--- a/src/gafaelfawr/models/kubernetes.py
+++ b/src/gafaelfawr/models/kubernetes.py
@@ -6,7 +6,7 @@ from abc import ABCMeta, abstractmethod
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from enum import Enum
-from typing import Optional
+from typing import Any, Literal, Optional
 
 from kubernetes_asyncio.client import (
     V1HTTPIngressPath,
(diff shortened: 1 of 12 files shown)


src/gafaelfawr/models/kubernetes.py
```diff
@@ -214,27 +222,82 @@ def scopes(self) -> list[str]:
...
+                msg = "authType: basic has no effect when loginRedirect is set"
+                raise ValueError(msg)
+
+        scopes = values.get("scopes")
+        if scopes and scopes.is_anonymous():
+            fields = ("auth_type", "delegate", "login_redirect", "replace_403")
+            for snake_name in fields:
+                if values.get(snake_name):
+                    camel_name = to_camel_case(snake_name)
+                    msg = f"{camel_name} has no effect for anonyous ingresses"
```
jonathansick, line 297:
```suggestion
                    msg = f"{camel_name} has no effect for anonymous ingresses"
```

jonathansick (approved):
Looks good. Looking forward to have a combo of anonymous ingresses and protected ingresses for Squareone.

PR #695, opened 2023-02-13
DM-37833: Add basic quota support

rra:
Support configuring default and per-group quotas in the Gafaelfawr settings, calculate the user's total quotas when user information is requested, and include that in the token user information model. Also reorder settings in gafaelfawr.config to match between settings and the frozen configuration and hopefully be a bit more readable.

Includes a bit of reformatting in the test suite that I noticed while updating the tests for quota handling.

### DIFF
--- a/src/gafaelfawr/config.py
+++ b/src/gafaelfawr/config.py
@@ -38,11 +38,14 @@ __all__ = [
     "GitHubSettings",
     "LDAPConfig",
     "LDAPSettings",
+    "NotebookQuota",
     "OIDCConfig",
     "OIDCClient",
     "OIDCServerConfig",
     "OIDCServerSettings",
     "OIDCSettings",
+    "Quota",
+    "QuotaGrant",
     "Settings",
 ]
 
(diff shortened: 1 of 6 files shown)


jonathansick (approved):
This looks good. The only hang-up I initially had was with the idea of "service names" and figuring out what name that actually was (a phalanx name? a URL path prefix?), and then from SQR-073 I realized it was the name of the Gafaelfawr ingress. In the end it makes sense, but I wonder if we'd want to replace "service names" with "ingress names" in the docs/comments to make it clearer? Say a service (in the sense of "App") employs two different ingresses, then it will have two different entries in the quota.

The 15 minute window is also a bit of a weird unit, but SQR-075 clarified that too.

rra:
The intent is that if the same service has multiple ingresses, they *would* all use the same service name, so that they would all count towards the same quota (unless, of course, we wanted to divide the service into multiple quotas, which Gafaelfawr represents by treating it as multiple services).

This service name is the same as the service name used with delegated tokens, which is recorded in the database entry for the delegated token as `service` as a record of which service the token was delegated to. Basically, there's a nascent identity management system for services here that isn't really fleshed out. Maybe I need to write a clear description of what this is. Normally, it's going to correspond to a Phalanx application name, although since it's the granularity of a quota, it's possible that we will need multiple services per application if they have to be quotaed differently.

PR #700, opened 2023-03-03
DM-38170: Disable caching of enrollment redirects

rra:
Some browsers appear to cache the 307 enrollment flow redirect and send the user back into the enrollment flow when they try to log in after they have enrolled.

Uniformly use Cache-Control: no-cache, no-store when disabling caching. (no-store should make no-cache unnecessary, but I believe it is a later keyword that earlier browsers may not support.) Previously, Gafaelfawr also added must-revalidate, but this doesn't appear to be necessary with modern browsers and was likely only meaningful combined with max-age=0 (which Gafaelfawr didn't set).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ Dependencies are updated to the latest available version during each release. Th
 
 ## 9.0.1 (unreleased)
 
+### Bug fixes
+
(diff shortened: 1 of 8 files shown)


jonathansick (approved):

jonathansick:
Did not know 307s might be cached; but definitely better safe.

PR #707, opened 2023-03-10
DM-38272: Improve logging and error reporting

rra:
Use the new Safir support for reporting uncaught exceptions, based on the earlier code in Gafaelfawr. Also use the new Safir `format_datetime_for_logging` instead of Gafaelfawr's version.

Make use of the Safir Slack client to report server-side errors during login to Slack. These include failures to verify upstream OpenID Connect tokens, failures to contact the authentication provider, and other problems that may indicate a system-wide outage or configuration issue rather than a problem with an individual login. Report those errors with 500 errors rather than 403 errors for slightly more accurate HTTP status codes.

Add the username to more log messages, particularly during user metadata retrieval during initial login.

Ensure that microseconds are stripped out of incoming expiration times rather than passing them to the database, which may or may not be able to store them.


rra:
On further thought, I don't think it's that useful to review this diff. It's kind of tedious and is mostly just improvements to exceptions and logging that are relatively obvious and not interestingly novel. I'm going to save reviewer energy for more interesting diffs.

PR #732, opened 2023-04-13
DM-38414: Switch to scriv for change log managmeent

rra:
Patching CHANGELOG.md in every PR causes unavoidable merge conflicts and rebase problems. Switch to scriv, which accumulates fragments of change log entries and only puts them into CHANGELOG.md during the release process.

Flesh out the documentation of how to maintain the change log, in addition to describing how to use scriv.


jonathansick (approved):
This looks really promising! I wonder if we ought to integrate `scriv create` and `scrib collect` into our Makefile to provide a bit more visibility into these commands. If you run `scriv collect` and leave out `--version` does it interactively ask for the version?

rra:
Sadly, no, it just creates an entry with no version number. Adding something to Makefile is good here, although I'm not sure I'd notice it there. We'd have to pick whether to add `--edit` or not, or provide two targets. (Users can configure this themselves in their Git config as well, but that's pretty obscure.)

PR #734, opened 2023-04-18
DM-38747: Support Kerberos GSSAPI binds to LDAP

rra:
Install the required libraries and the k5start program in the container and, if a keytab and krb5.conf file are present in the container, run the application under k5start. (This will be handled by the Helm chart for invocations other than the default uvicorn one.) If ldap.useKerberos is set in the configuration and there is no userDn set, use GSSAPI as the authentication method.

Add documentation for the secret and Helm configuration. Reorganize the LDAP configuration documentation a bit to use subheadings.


athornton (approved):
LGTM.  I mean, I have no idea what the right k5start flags are but I'm assuming you know what you're doing there.  Very cool to have this capability.

PR #984, opened 2024-03-13
DM-43288: Add a health check endpoint

rra:
Add a new internal endpoint, /health, which checks the database, Redis, and (if configured) LDAP and raises an uncaught exception if any of them fail. This will be used as a Kubernetes liveness check to address the issues we've had with Gafaelfawr recovering from Redis or LDAP outages and losing its database connections.


rra:
This seemed worth a code review, not because the code is all that complex, but because I wasn't sure if you had any thoughts about the best way to structure health checks like this. We'll eventually want to add these to more applications (and in some cases will need to implement the IVOA availability endpoint).

jonathansick (approved):
This seems good. I don't have any ideas on how to do it differently, so I think we can go with this.

PR #1102, opened 2024-09-16
fix: typo in gafaelfawringress.rst

ac6y:


rra:
Thank you for the fix!

PR #1246, opened 2025-02-25
DM-49151: Switch to uv lock files

rra:
Stop generating traditional frozen dependency files and instead use a uv lock file with dependency groups configured in `pyproject.toml`. Update the tox, Docker, and GitHub Actions configuration accordingly.

For now, avoid the `lsst-sqre/run-tox` action and run tox directly. This support will, once tested, be rolled into `lsst-sqre/run-tox`.

Stop linting the JavaScript UI in pre-commit, since this requires a lot of manual maintenance and is essentially never relevant. Instead, run the lint step only in GitHub Actions. Add pre-commit configuration for uv to ensure `uv.lock` is up to date.

Add a `.python-version` file and use that to determine what version of Python to use in GitHub Actions.

Rewrite the `Dockerfile` to make use of cache file systems and use the recommended mechanism of uv integration.

Rework the tox configuration to use dependency groups and other features of tox-uv to try to optimize the tox virtualenv management.


rra:
The easiest way to review this is probably to look at the new `Makefile`, `tox.ini`, `Dockerfile`, and GitHub Actions configuration from the branch, since the diffs are pretty messy because of things moving around.

Dockerfile
```diff
@@ -16,67 +16,60 @@ FROM python:3.13.2-slim-bookworm AS base-image
...
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --frozen --compile-bytecode --no-install-project
 
 # Install the Gafaelfawr Python application.
-COPY . /workdir
-WORKDIR /workdir
-RUN uv pip install --compile-bytecode --no-cache .
+ADD . /app
+RUN --mount=type=cache,target=/.root/.cache/uv \
+    uv pip install --compile-bytecode .
```
jonathansick, line 51: Would a `--no-deps` flag be a good idea here since we did a `uv sync` earlier?

Dockerfile
```diff
@@ -16,67 +16,60 @@ FROM python:3.13.2-slim-bookworm AS base-image
...
-# Install the app's Python runtime dependencies
-COPY requirements/main.txt ./requirements.txt
-RUN uv pip install --compile-bytecode --verify-hashes --no-cache \
-    -r requirements.txt
+# Install the dependencies.
+WORKDIR /app
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --frozen --compile-bytecode --no-install-project
```
jonathansick, line 46: Do you need a `--no-dev` flag here to ensure the dev group doesn't get included automatically? (does it?)

docs/dev/development.rst
```diff
@@ -21,41 +21,68 @@ Gafaelfawr is developed by the LSST SQuaRE team.
...
 Otherwise, OpenSSL isn't on the default linker path and some Python extensions may not build.
+This must be set in any terminal window where you may run :command:`tox`.
+
+Finally, you can optionally enter the Gafaelfawr development virtualenv with:
+
+.. prompt:: bash
+
+   source .venv/bin/activate
+
+This is optional; you do not have to activate the virtualenv to do development.
```
jonathansick, line 67: I think VS Code struggles to do its intellisense magic if an environment isn't available with the app in it, but I get the sentiment here.

jonathansick (approved):
This is pretty exciting. The main question I have is related to the CI use case where we're doing a `uv sync` followed by a `uv run`. I wonder if we can drop the `uv sync` and just particular about what dependency groups are involved in the `uv run`?

I also had a thought about the `uv sync` in the docker image and whether that accidentally brings in dev dependencies? I'm not sure  if `dev` is treated as a default group that's always sync'd.

.github/workflows/ci.yaml
```diff
@@ -95,97 +82,141 @@ jobs:
...
         uses: medyagh/setup-minikube@v0.0.19
         with:
           kubernetes-version: "v1.27.3"
 
-      - uses: lsst-sqre/run-tox@v1
+      - name: Install tox dependencies
+        run: uv sync --only-group tox
+
+      - name: Run tox
+        run: uv run tox run -e lint,typing,py-full,coverage-report
```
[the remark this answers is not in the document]
rra: You are entirely correct. Thank you for catching that! This is way better. I missed that `uv run` itself did a sync.

Dockerfile
rra: I think it shouldn't make a difference since it should immediately recognize the dependencies are satisfied, but it certainly wouldn't hurt. Done.

Dockerfile
rra: I do, yes, thank you! `--no-default-groups` in particular.

docs/dev/development.rst
rra: I added a note saying that editors with Python integration such as VSCode may work better if you activate the virtualenv before starting them.

PR #1254, opened 2025-03-05
DM-49315: Add support for per-user domain authorization

rra:
Add a new setting to `GafaelfawrIngress` named `config.userDomain` that, if set to true, requires that the last component of the request URL match the authenticated username. This is intended for use with per-user subdomains, such as Jupyter labs with JavaScript origin isolation. In this case, as an additional sanity check, also require that the hostname in the request be a subdomain of Gafaelfawr's base domain.


rra:
This branch was broken into separate PRs and merged piecemeal.

PR #1425, opened 2026-03-13
Add `vinylAuthCacheEnabled` config option to `GafaelfawrIngress` CRD

fajpunk:
If this is set to true, then all auth requests for a given ingress will be sent to the the Vinyl cache sidecar container in the Gafaelfawr pod. This cache is configured to cache all GET responses by Authorization header and cookie for five minutes.


fajpunk:
This was used for auth cache performance testing:
https://github.com/lsst-sqre/sqr-111/pull/4

PR #1457, opened 2026-05-27
DM-55072: Add IVOA GMS protocol implementation

rra:
Add an implementation of the IVOA GMS 1.0 protocol on the `/auth/gms` route. This is a simple protocol that allows IVOA services to determine the group membership of the current authenticated user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Read those documents for a more complete picture of how Gafaelfawr fits into a l
 Gafaelfawr provides authentication and access control via NGINX's `auth_request` directive, and handles integration with an external identity provider (either with GitHub or OpenID Connect).
 Authentication sessions and user identity information are stored in Redis.
 Token information is stored in a SQL database.
-It also provides an API to create and manipulate tokens, and a minimal OpenID Connect server to support protected services that only understand OpenID Connect.
+It also provides basic API rate limiting and user quota information, an API to create and manipulate tokens, a minimal OpenID Connect server to support protected services that only understand OpenID Connect, and an implementation of the IVAO Group Membership Service protocol (version 1.0).
 
 For full documentation, see [gafaelfawr.lsst.io](https://gafaelfawr.lsst.io/).
 
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Read those documents for a more complete picture of how Gafaelfawr fits into a l
 Gafaelfawr provides authentication and access control via NGINX's `auth_request` directive, and handles integration with an external identity provider (either with GitHub or OpenID Connect).
 Authentication sessions and user identity information are stored in Redis.
 Token information is stored in a SQL database.
-It also provides basic API rate limiting and user quota information, an API to create and manipulate tokens, a minimal OpenID Connect server to support protected services that only understand OpenID Connect, and an implementation of the IVAO Group Membership Service protocol (version 1.0).
+It also provides basic API rate limiting and user quota information, an API to create and manipulate tokens, a minimal OpenID Connect server to support protected services that only understand OpenID Connect, and an implementation of the IVOA Group Membership Service protocol (version 1.0).
 
 For full documentation, see [gafaelfawr.lsst.io](https://gafaelfawr.lsst.io/).
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -16,8 +16,10 @@ Gafaelfawr
 Gafaelfawr provides the authentication and authorization infrastructure for Phalanx_ environments, including the Vera C. Rubin Observatory Science Platform.
 
 Its primary purpose is to serve as an NGINX ``auth_request`` backend.
-It also provides a web page where people can create and manage long-lived tokens for use outside of a web browser, and can serve as a simple OpenID Connect server.
-Gafaelfawr requires the Kubernetes `NGINX ingress controller <https://github.com/kubernetes/ingress-nginx>`__.
+It also provides basic API rate limiting and user quota information, an API to create and manipulate tokens, a minimal OpenID Connect server to support protected services that only understand OpenID Connect, and an implementation of the `IVAO Group Membership Service protocol (version 1.0) <https://www.ivoa.net/documents/GMS/20220222/REC-GMS-1.0.html>`__.
+
+Currently, the Kubernetes `NGINX ingress controller <https://github.com/kubernetes/ingress-nginx>`__.
+A future version will use a Kubernetes gateway controller (probably Envoy) instead.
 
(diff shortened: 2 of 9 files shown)


src/gafaelfawr/handlers/gms.py
```diff
@@ -0,0 +1,84 @@
...
+
+    # Construct the response. If no groups are provided, list all of the
+    # user's groups. If groups are provided, only list the intersection
+    # between that list of groups and the user's groups. Each group line must
+    # end with a newline, but if there are no matching groups, the body must
+    # be completely empty.
+    seen = {g.name for g in user_info.groups}
+    matching = seen & wanted if wanted else seen
+    if matching:
+        return "\n".join(sorted(matching)) + "\n"
```
stvoutsin: Looking at the spec it seems they want "\r\n" line endings:
"Each group name (even the last) must end with a newline as a CRLF2"

README.md
```diff
@@ -11,7 +11,7 @@ Read those documents for a more complete picture of how Gafaelfawr fits into a l
 Gafaelfawr provides authentication and access control via NGINX's `auth_request` directive, and handles integration with an external identity provider (either with GitHub or OpenID Connect).
 Authentication sessions and user identity information are stored in Redis.
 Token information is stored in a SQL database.
-It also provides an API to create and manipulate tokens, and a minimal OpenID Connect server to support protected services that only understand OpenID Connect.
+It also provides basic API rate limiting and user quota information, an API to create and manipulate tokens, a minimal OpenID Connect server to support protected services that only understand OpenID Connect, and an implementation of the IVAO Group Membership Service protocol (version 1.0).
```
stvoutsin:
```suggestion
It also provides basic API rate limiting and user quota information, an API to create and manipulate tokens, a minimal OpenID Connect server to support protected services that only understand OpenID Connect, and an implementation of the IVOA Group Membership Service protocol (version 1.0).
```

docs/user-guide/gms.rst
```diff
@@ -0,0 +1,27 @@
+:og:description: Learn how to use the Gafaelfawr IVOA GMS API.
+
+.. _gms:
+
+################
+IVOA GMS queries
+################
+
+Gafaelfawr also provides an implementation of the `IVAO Group Membership Service protocol (version 1.0) <https://www.ivoa.net/documents/GMS/20220222/REC-GMS-1.0.html>`__.
```
stvoutsin:
```suggestion
Gafaelfawr also provides an implementation of the `IVOA Group Membership Service protocol (version 1.0) <https://www.ivoa.net/documents/GMS/20220222/REC-GMS-1.0.html>`__.
```

docs/index.rst
```diff
@@ -16,8 +16,10 @@ Gafaelfawr
 Gafaelfawr provides the authentication and authorization infrastructure for Phalanx_ environments, including the Vera C. Rubin Observatory Science Platform.
 
 Its primary purpose is to serve as an NGINX ``auth_request`` backend.
-It also provides a web page where people can create and manage long-lived tokens for use outside of a web browser, and can serve as a simple OpenID Connect server.
-Gafaelfawr requires the Kubernetes `NGINX ingress controller <https://github.com/kubernetes/ingress-nginx>`__.
+It also provides basic API rate limiting and user quota information, an API to create and manipulate tokens, a minimal OpenID Connect server to support protected services that only understand OpenID Connect, and an implementation of the `IVAO Group Membership Service protocol (version 1.0) <https://www.ivoa.net/documents/GMS/20220222/REC-GMS-1.0.html>`__.
```
stvoutsin:
```suggestion
It also provides basic API rate limiting and user quota information, an API to create and manipulate tokens, a minimal OpenID Connect server to support protected services that only understand OpenID Connect, and an implementation of the `IVOA Group Membership Service protocol (version 1.0) <https://www.ivoa.net/documents/GMS/20220222/REC-GMS-1.0.html>`__.
```

src/gafaelfawr/handlers/gms.py
```diff
@@ -0,0 +1,84 @@
...
+__all__ = ["router"]
+
+
+@router.get(
+    "/auth/gms",
+    description=(
+        "Get group information in IVOA GMS format for the authenticated user"
+    ),
+    response_class=PlainTextResponse,
+    responses={401: {"description": "Unauthenticated"}},
```
stvoutsin, line 30: Do we want to add 200 and 500 here?

src/gafaelfawr/handlers/gms.py
```diff
@@ -0,0 +1,84 @@
...
+    summary="Get user groups",
+    tags=["user"],
+)
+async def get_gms(
+    group: Annotated[list[str] | None, Query()] = None,
+    *,
+    request: Request,
```
stvoutsin: Can request be removed here since I don't see it being used?

src/gafaelfawr/handlers/gms.py
```diff
@@ -0,0 +1,84 @@
...
+    # Construct the response. If no groups are provided, list all of the
+    # user's groups. If groups are provided, only list the intersection
+    # between that list of groups and the user's groups. Each group line must
+    # end with a newline, but if there are no matching groups, the body must
+    # be completely empty.
+    seen = {g.name for g in user_info.groups}
+    matching = seen & wanted if wanted else seen
+    if matching:
+        return "\n".join(sorted(matching)) + "\n"
+    else:
```
stvoutsin, line 74: Could just remove the else here if you want

stvoutsin (approved):
This all looks great to me. Only a couple of typos and and minor comments

src/gafaelfawr/handlers/gms.py
rra: 200 has a description by default since it's the default return status and 500 is a "this should never happen case" about which we can't really say anything definitive, so I usually leave it out. This is used in the API documentation and 5xx errors are somewhat by definition "unexpected" rather than documentable.

src/gafaelfawr/handlers/gms.py
rra: Yeah, this was left over from debugging, thanks, good catch.

src/gafaelfawr/handlers/gms.py
rra: Good catch, I thought FastAPI and HTTPX were doing automatic line ending translation for some reason but of course they don't. Fixed.

src/gafaelfawr/handlers/gms.py
rra: Yeah, indeed, and I thought about it, but it feels a bit clearer that there are two explicit cases.

